PR #57, opened 2015-02-21
Remove impossible code for 'completed' event

pushrax:
This code used to work since we could tell if a peer was present in both IP families. We don't have a way to do this now. However, since ba1ad7f5bfadaee6043642bb363c6d7468162a74, it doesn't matter, since line 239 would be hit if `v4seed` or `v6seed` were true and the whole announce will error out.

@jzelinskie does this seem right?


jzelinskie:
@pushrax lgtm

Issue #59, opened 2015-03-20
Clients with the same IP address, but different ports are not considered separate peers


jzelinskie:
Was trying to setup a demo using docker containers, which internally use the same IP but different ports. Our `peerkey`s don't take into account that peers could be from the same IP address, but be operating on different ports.

PR #61, opened 2015-03-22
Add Port to PeerKey

jzelinskie:
This makes clients with the same IP address, but different ports distinguishable.
I'm pretty sure this doesn't require any changes for PeerMaps.

PTAL


jzelinskie:
btw this is meant to fix #59

pushrax:
:cool:

PR #68, opened 2015-06-27
logging: add error to udp logs

jzelinskie:
This also makes the format of UDP and HTTP logs a little more similar.

### DIFF
--- a/udp/protocol.go
+++ b/udp/protocol.go
@@ -62,9 +62,12 @@ func handleTorrentError(err error, w *Writer) {
 }
 
 // handlePacket decodes and processes one UDP request, returning the response.
-func (s *Server) handlePacket(packet []byte, addr *net.UDPAddr) (response []byte, actionName string) {
+func (s *Server) handlePacket(packet []byte, addr *net.UDPAddr) (response []byte, actionName string, err error) {
 	if len(packet) < 16 {
-		return // Malformed, no client packets are less than 16 bytes.
+		// Malformed, no client packets are less than 16 bytes.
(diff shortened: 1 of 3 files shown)


udp/udp.go
```diff
@@ -68,19 +68,22 @@ func (s *Server) serve(listenAddr string) error {
...
 		}
 
-		start := time.Now()
-
 		go func() {
-			response, action := s.handlePacket(buffer[:n], addr)
-			pool.GiveSlice(buffer)
+			start := time.Now()
+			response, action, err := s.handlePacket(buffer[:n], addr)
+			defer pool.GiveSlice(buffer)
```
pushrax: Any particular reason to defer this?

udp/protocol.go
```diff
@@ -96,8 +102,9 @@ func (s *Server) handlePacket(packet []byte, addr *net.UDPAddr) (response []byte
 
 	case announceActionID:
 		actionName = "announce"
-		ann, err := s.newAnnounce(packet, addr.IP)
 
+		var ann *models.Announce
+		ann, err = s.newAnnounce(packet, addr.IP)
```
pushrax: There needs to be a better way of doing this in Go

udp/udp.go
jzelinskie: I didn't really want it to count in the timing, but wanted to keep it in the context of the `handlePacket` call.

udp/protocol.go
jzelinskie: Let me know when you figure it out :\

PR #69, opened 2015-06-27
tracker: rm badrequest when state is unknown

jzelinskie:
This assumption was just totally wrong. If there is any kind of packet
loss or the tracker starts up to a pre-existing swarm, all the peers
will get bad request responses.


pushrax:
Thanks for finding this :+1:

PR #74, opened 2015-06-30
udp: make BadConnID a normal error

### LINKED ISSUE #72: BadConnectionID errors are returned if the tracker is shutdown
The tracker starts up and generates a random key to use for the generation of connection IDs for UDP packets, but if for some reason the tracker restarts or, if in the future, when clustering is available, traffic gets routed to another tracker, these keys will be different and clients will be served a "BadConnectionID" error.

This can be resolved by adding a key into the config and simply using that.

jzelinskie:
This causes packets that have bad connection IDs to not be recorded as a
bad request in our stats. These types of errors to be expected and
should balloon our stats.

Fixes #72.

### DIFF
--- a/udp/protocol.go
+++ b/udp/protocol.go
@@ -7,6 +7,7 @@ package udp
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"net"
 
 	"github.com/chihaya/chihaya/stats"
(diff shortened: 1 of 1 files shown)


dtrackd:
:thumbsup:

PR #75, opened 2015-07-01
Fix Dockerfile to build with the latest changes

pushrax:
Things have changed a bit, which broke the Docker build. I also moved dependency fetching to before source addition to prevent the dependency cache from being busted when Chihaya source changes.

@jzelinskie


jzelinskie:
I was literally about to make this fix. If you could amend the commit message to do the normal format, LGTM.

PR #76, opened 2015-07-03
chihaya: refactor ALL the things

jzelinskie:
This reintroduces an interface for the primary storage of the tracker
and names it "Store". It also renames "backend" to "DeltaStore" which is
a less ambiguous name. This also moves transport implementations into a
transport directory.

I'm sorry for the huge monolithic commit (refactoring slowly brings in everything) and sorry for the millionth rename of the "storage" stuff.


jzelinskie:
@pushrax any comments so far?

film42:
I've never contributed to this project before, but found it after chatting on IRC. I was reading through this PR and thought I might comment.

I like the idea of separating the driver into a producer and consumer, but I'm not sure how useful it becomes to the developer adding their own driver. This comes from no experience working with the software, but it seems early on to start breaking apart an already very simple interface.

Also, did you mean to create the new `no-op` driver `noop`? Currently it's named `nop`.

Everything else looks good :)

jzelinskie:
We had a little conversation on IRC surrounding this PR (#chihaya on freenode). This PR is still up in the air. Because we really don't have anything driving the development of the producer/consumer, the design is largely speculative.

I believe I changed `noop` to `nop` because I saw `nop` being used elsewhere in the Go standard library and I like to try and keep things as consistent and idiomatic as possible.

film42:
I didn't know `nop` was the terminology being used in the std lib, TIL.

I think it might be better to leave out producer/ consumer stuff for now but merge in the other changes. I think once more developers start adding drivers, you'll be able to include a lot of empirical evidence into the design (my opinion).

PR #80, opened 2015-07-13
Fix port overload

klimkin:
I have discovered, that tracker didn't work with ctorrent when using ctorrent with default parameters. Tracker actually worked when IP was forced in ctorrent, so it would send "ip" parameter together with "port". After some debug, I found, that tracker was using port from HTTP request source IP, which was not the port the client was listening to.
After the fix ctorrent was able to work properly.

### DIFF
--- a/example_config.json
+++ b/example_config.json
@@ -8,7 +8,7 @@
   "default_num_want": 50,
   "torrent_map_shards": 1,
   "allow_ip_spoofing": true,
-  "dual_stacked_peers": false,
+  "dual_stacked_peers": true,
   "real_ip_header": "",
   "respect_af": false,
   "client_whitelist_enabled": false,
--- a/example_config.json
+++ b/example_config.json
@@ -8,7 +8,7 @@
   "default_num_want": 50,
   "torrent_map_shards": 1,
   "allow_ip_spoofing": true,
-  "dual_stacked_peers": true,
+  "dual_stacked_peers": false,
   "real_ip_header": "",
   "respect_af": false,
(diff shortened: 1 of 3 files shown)


jzelinskie:
I think I see the bug here. We fallback to getting data from the `RemoteAddr`, which has local port information. I think the better solution is to attempt to strip the port from `RemoteAddr` before passing it to `getEndpoints` because getEndpoints will handle the zeroing out of the port.

pushrax:
Actually I think the correct solution is to completely ignore any port specified in any IP parameters. afaik the IP parameter shouldn't ever have the port included.

jzelinskie:
sgtm, I can do the PR unless @klimkin wants to get more familiar with the codebase.

klimkin:
I have reverted the change to the default config, so we can handle it separately. Two things:

1. It make sense to keep defaults and example in sync
2. The dual-stacked-ips feature is confusing. When true, it actually keeps trying to parse IP using getEndpoints if previous call didn't provide both ipv4 and ipv6. So, in requestedEndpoint, even if IP was provided with "ip" parameter, it continues with r.RemoteAddr, overriding "spoofed" IP. Looks like a bug to me.

klimkin:
The last commit failed CI. Did it lose UDP packet? :)

cooperlees:
Should we add the code to disable this if we set respect_af - I need to read the actual code around this tho ...

jzelinskie:
I think this was fixed in #83. I'll reopen if it isn't.

PR #82, opened 2015-08-19
travis: update to Go 1.5

jzelinskie:
No idea if this is going to pass the builds, but this is probably what we should target.


jzelinskie:
I actually don't think the bug is in this PR. I think one of the UDP tests are broken.

PR #83, opened 2015-08-20
http: stop using remote http port

jzelinskie:
The remote port used to send the announce has nothing to do with the
port being listened to for the torrent client protocol.

### DIFF
--- a/http/tracker.go
+++ b/http/tracker.go
@@ -44,31 +44,17 @@ func (s *Server) newAnnounce(r *http.Request, p httprouter.Params) (*models.Anno
 	}
 
 	left, err := q.Uint64("left")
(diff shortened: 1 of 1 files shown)


http/tracker.go
```diff
@@ -132,26 +118,26 @@ func requestedPeerCount(q *query.Query, fallback int) int {
 	return fallback
 }
 
-// requestedEndpoint returns the IP address and port pairs for a request. If
+// requestedIP returns the IP address and port pairs for a request. If
```
jzelinskie: I need to remove `and port pairs` from this doc string.

cooperlees:
LGTM

PR #86, opened 2015-08-26
tracker: fix logic recording complete events

jzelinskie:
520a357 inadvertently changed the logic on how many times
leecherFinished was being called and recording completion events to the
stats. This commit makes this that clearer and avoids over incrementing
the number of Seeders in our stats.


cooperlees:
LGTM - Many thanks for this sir!

Issue #65, opened 2015-06-18
Potential Memory Leak


dtrackd:
Working from develop brach on a debain 8 64bit system
go version go1.4.2 linux/amd64

config: 
```
{
  "create_on_announce": true,
  "private_enabled": false,
  "freeleech_enabled": false,
  "purge_inactive_torrents": false,
  "announce": "30m",
  "min_announce": "15m",
  "default_num_want": 50,
  "torrent_map_shards": 1,
  "allow_ip_spoofing": true,
  "dual_stacked_peers": true,
  "real_ip_header": "",
  "respect_af": false,
  "client_whitelist_enabled": false,
  "client_whitelist": ["OP1011"],
  "udp_listen_addr": ":80",
  "http_listen_addr": ":8080",
  "http_request_timeout": "4s",
  "http_read_timeout": "4s",
  "http_write_timeout": "4s",
  "http_listen_limit": 0,
  "driver": "noop",
  "stats_buffer_size": 0,
  "include_mem_stats": true,
  "verbose_mem_stats": false,
  "mem_stats_interval": "5s"
}
```

Running the tracker seems to make the memeroy usage continue to grow at a pretty constant rate of ~3MB every 5 seconds or so under real world traffic.

the server has 1GB of memory and no other daemons running, its only used for chihaya.
So in reality we are able to play with ~850MB of memory.

Here you can also see the growth in ram over periods of time (`4156` is the PID of the tracker)

```
root@box:~/chihaya/bin# date; cat /proc/4156/status | grep VmSize
Fri Jun 19 01:30:43 CEST 2015
VmSize:	  668200 kB
```

(~8mins later...)
```
root@box:~/chihaya/bin# date; cat /proc/4156/status | grep VmSize
Fri Jun 19 01:38:32 CEST 2015
VmSize:	  708752 kB
```

(~4mins mins after that... )
```
root@box:~/chihaya/bin# date; cat /proc/4156/status | grep VmSize
Fri Jun 19 01:42:40 CEST 2015
VmSize:	  737320 kB
```

At the point of when the tracker was using `737320 kB` of memory the stats showed as follows:

```json
Started: "2015-06-19T01:09:20.30660432+02:00",
Connections.Open: 1,
Connections.Accepted: 66,
BytesTransmitted: 0,
Runtime.GoRoutines: 17,
Requests.Handled: 1,
Requests.Errored: 0,
Requests.Bad: 2625430,
ResponseTime: {
P50: 0.236624,
[...]

PR #89, opened 2015-09-18
config: make reaping configurable

jzelinskie:
This allows for configurable reaping rates and changes the default to
what many public trackers use in the wild.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -81,6 +81,8 @@ type TrackerConfig struct {
 	PurgeInactiveTorrents bool     `json:"purge_inactive_torrents"`
 	Announce              Duration `json:"announce"`
 	MinAnnounce           Duration `json:"min_announce"`
+	ReapInterval          Duration `json:"reap_interval"`
+	ReapRatio             float64  `json:"reap_ratio"`
 	NumWantFallback       int      `json:"default_num_want"`
 	TorrentMapShards      int      `json:"torrent_map_shards"`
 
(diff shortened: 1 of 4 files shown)


jzelinskie:
@rarbg would you like to test this branch?

pushrax:
It seems like this would reduce memory consumption by at most a factor of 2 (2.5/1.25); I can't rationalize how it fixes the behaviour reported in #65.

rarbg:
@pushrax 
You didnt read the commit 
Its not only changing the time after which peers are removed , but it changes the frequency on which the gc is ran
The problem was the mistake with linear vs exponential growth.

pushrax:
If peers are reaped every T, and removed after not announcing for 2T, then the longest a peer can possibly remain after they stop announcing is 2T+T. Under the scheme in this commit, that duration is reduced to about 1.25T – a constant factor. I don't see the exponential growth.

jzelinskie:
This PR just allowed more customization of the reaping of peers which is the primary way memory pressure builds. I don't particularly see any issue with it.

When I ran the tracker and made it reap all peers older than a second and GC every second, memory never grew, so there is no memory leak. Rather, people are just confused because of the rate at which we GC and memory builds up.

If you want to change the defaults -- I'm totally okay with that. I just used what they said because they claimed that's pretty standard for how public trackers are run.

rarbg:
@pushrax
"If peers are reaped every T, and removed after not announcing for 2T"
The commit introduces peers are reaped every Y and removed after not announcing for XT
@pushrax @jzelinskie 
Let me be perfectly clear...
1. It doesnt fix the high memory usage - 50mln peers = 74g ram
2. It does prevent the memory spiking because the previous code before that commit didnt clean peers at the appropriate intervals.
3. If a dmca agency hits your tracker with 10 mln announces - you can calculate yourself what happens with previous code vs this code

pushrax:
@jzelinskie no issue at all, I just couldn't figure out how it would result in an order of magnitude difference (which it's now clear it doesn't). :+1: 

@rarbg right, I was speaking about the earlier behaviour, justifying that "factor of 2.4" statement I made earlier. I think that factor is still correct – this commit should reduce memory spikes by a factor of at most 2.4.

PR #111, opened 2015-10-28
backend: remove unused parameter from LoadAllUsers()

### LINKED ISSUE #107: LoadAllUsers should have no parameters
In [backend/backend.go:74](https://github.com/chihaya/chihaya/blob/master/backend/backend.go#L74): `LoadAllUsers([]uint64)` should have no parameters, as it loads all users.

mrd0ll4r:
Fixes #107 

Note: The gazelle driver will have to be updated as well to implement `backend.Conn`, will do that in a minute

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -71,5 +71,5 @@ type Conn interface {
 	LoadUsers(ids []uint64) ([]*models.User, error)
 
 	// LoadAllUsers fetches and returns all users.
-	LoadAllUsers(ids []uint64) ([]*models.User, error)
+	LoadAllUsers() ([]*models.User, error)
 }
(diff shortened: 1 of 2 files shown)


jzelinskie:
The gazelle driver is unmaintained. Personally, I've never used Gazelle, so I don't have the knowledge to know how to best integrate it. It also looks like the Gazelle guys are attempting to rewrite their stack so I'd rather target the next-gen Gazelle if the rewrite becomes usable.

This is why I'm debating just ripping out the backend driver until there are much clearer specifications for what needs to be done.

mrd0ll4r:
Oh my, yes, I've just realized that the gazelle driver, even at its current stage, does not work.
I can update it to theoretically work again (mostly), but I see your point here - let's wait for the gazelle rewrite ;-)

PR #112, opened 2015-10-28
tracker: update reslicing to use implicit slice bounds

### LINKED ISSUE #94: Implicit slice bounds not used when reslicing
In [tracker/models/models.go](https://github.com/chihaya/chihaya/blob/master/tracker/models/models.go#L178) the implicit slice bounds (in this case the `0`) are not used - this could be `a.PeerID[:6]`. This is so far the only case I've found, but we could just let `go fix` run over all the sources to find possible other instances of this.

mrd0ll4r:
Fixes #94 

Unfortunately we can not just `go fix` the whole project - I found out that `go fix` only fixes new features for the latest release. Using the `fix` from the go version where implicit slice bounds were introduced will probably not work (because it's ancient).

### DIFF
--- a/tracker/models/models.go
+++ b/tracker/models/models.go
@@ -175,7 +175,7 @@ func (a *Announce) ClientID() (clientID string) {
 				clientID = a.PeerID[1:7]
 			}
 		} else {
-			clientID = a.PeerID[0:6]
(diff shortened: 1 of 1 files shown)


jzelinskie:
LGTM

PR #114, opened 2015-10-28
udp: replace boolean-synchronization with channels

### LINKED ISSUE #108: closing/synchronization in udp should be done using channels
The `done bool` field of `udp.Server` should be replaced with a `close(d)/done chan struct{}` that is closed to shut down the server. (and possibly a `WaitGroup`)

This also allows us to properly kill the goroutine that updates the IV ([udp/udp.go:98-106](https://github.com/chihaya/chihaya/blob/master/udp/udp.go#L98-L106)) with a `select` between the timer and the closed chan.

mrd0ll4r:
Fixes #108 

I'm pretty sure this works, but there are no tests checking that, are there?

### DIFF
--- a/udp/udp.go
+++ b/udp/udp.go
@@ -22,7 +22,7 @@ import (
 type Server struct {
 	config  *config.Config
 	tracker *tracker.Tracker
-	done    bool
+	closing chan struct{}
 	booting chan struct{}
 	sock    *net.UDPConn
 
@@ -55,7 +55,12 @@ func (s *Server) serve() error {
 	s.sock = sock
 	close(s.booting)
 
-	for !s.done {
+	for {
+		select {
+		case <-s.closing:
+			return nil
+		default:
+		}
 		buffer := pool.TakeSlice()
 		sock.SetReadDeadline(time.Now().Add(time.Second))
 		n, addr, err := sock.ReadFromUDP(buffer)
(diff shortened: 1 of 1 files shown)


jzelinskie:
Why not use a waitgroup? It barely changes any of the code.

```diff
diff --git a/udp/udp.go b/udp/udp.go
index fdc0b78..b96ae9e 100644
--- a/udp/udp.go
+++ b/udp/udp.go
@@ -9,6 +9,7 @@ package udp
 import (
        "errors"
        "net"
+       "sync"
        "time"

        "github.com/golang/glog"
@@ -22,9 +23,10 @@ import (
 type Server struct {
        config  *config.Config
        tracker *tracker.Tracker
+       sock    *net.UDPConn
        done    bool
+       wg      sync.WaitGroup
        booting chan struct{}
-       sock    *net.UDPConn

        connIDGen *ConnectionIDGenerator
 }
@@ -68,7 +70,9 @@ func (s *Server) serve() error {
                        return err
                }

+               s.wg.Add(1)
                go func() {
+                       defer s.wg.Done()
                        start := time.Now()
                        response, action, err := s.handlePacket(buffer[:n], addr)
                        defer pool.GiveSlice(buffer)
@@ -116,6 +120,7 @@ func (s *Server) Serve() {
 func (s *Server) Stop() {
        s.done = true
        s.sock.SetReadDeadline(time.Now())
+       s.wg.Wait()
 }

 // NewServer returns a new UDP server for a given configuration and tracker.
```

jzelinskie:
Upon looking at this change. Both are required. We need the closing channel to signal to every process that we're going to be done and we need to waitgroup to know when we are actually done.

mrd0ll4r:
Yes that's right, I forgot to add the waitgroups in the first commit, will do that.

Issue #53, opened 2015-02-16
UDP support


pushrax:
Should be relatively simple to add by implementing a `tracker.Writer` for UDP.

jzelinskie:
http://www.bittorrent.org/beps/bep_0007.html

Leaving this link here as it is relevant for the IPv6 UDP support.

jzelinskie:
...and here's the actual UDP spec
http://www.bittorrent.org/beps/bep_0015.html

pushrax:
UDP support has been merged into `develop`, though there are still probably some issues present. Leaving this open until someone is running UDP in production successfully.

jzelinskie:
This got merged into master

PR #118, opened 2015-11-01
udp: parse URLData according to BEP41

mrd0ll4r:
This is WIP.

I added a first version of URLData optional parameter parsing for UDP announces, as described in [BEP41](http://bittorrent.org/beps/bep_0041.html).

Additionally, I fixed parsing of optional parameters in general, there was a problem with indices, whenever URLData was present and followed by more optional parameters. (see l. 224 in the updated protocol.go)

Now, the only thing I extract from the URLData is the passkey. The problem is: URLData could contain way more than that, but I don't know if we can store that in the announce somehow, or even if we should (think someone sending us bogus parameter strings, we store them in the announce and try to handle them).

Any suggestions? What else could I parse? How to handle more query parameters, e.g. `"/path/something?passkey=abc123"` - what to do with `path` and `something`?

**EDIT**: This somewhat contributes to #53

### DIFF
--- a/udp/protocol.go
+++ b/udp/protocol.go
@@ -12,6 +12,8 @@ import (
 
 	"github.com/chihaya/chihaya/stats"
 	"github.com/chihaya/chihaya/tracker/models"
+	"net/url"
+	"strings"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


mrd0ll4r:
Will close this for the time being, because of the tracker overhaul

Issue #44, opened 2015-02-08
Improve logging w/ standard levels


jzelinskie:
The current selection of log levels confused me. When `-v=0`, our "purging peers" is displayed, but not requests?

I think we need to come up with guidelines for our log lines and start sprinkling them _everywhere_.

mrd0ll4r:
I think we should introduce an interface for a logger, something like

    type Logger interface {
        New(name string) Logger

    	Errorln(message string) //or maybe Fatalln, Fatal
    	Error(format string, a ...interface{})
    
    	Warnln(message string)
    	Warn(format string, a ...interface{})
    
    	Infoln(message string)
    	Info(format string, a ...interface{})
    
    	Debugln(message string)
    	Debug(format string, a ...interface{})
    
    	Traceln(message string)
    	Trace(format string, a ...interface{})
    }

Maybe we could even leave out the `*ln` functions...

Especially, I would want to include checks like `if glog.V(2)` in the respective functions (in this case `Infoln` and `Info`).

That allows us to just write something like `log.Debug("Reaped %d inactive torrents",reaped)` instead of the three-line `if` construct...

The `New` function allows us to name loggers by package, I guess that could be nice. Maybe we should do something like the syncthing guys, specifying an environmental variable defining which packages should log on debug level, or specify that by flags.

What do you think?

Just a sidenote: Why do we use `glog`?

jzelinskie:
What your suggesting looks a lot like [capnslog](github.com/coreos/capnslog). Creating a logger for each package gets pretty annoying, especially when glog already points out the line in the file anyway. I'm not against leveled logging; in fact, that's why we originally chose glog. The `if` construct is very powerful and allows us to do things like [this](https://github.com/chihaya/chihaya/blob/b303f0e9f2883505322d38e7dc874ed6607cabc8/http/http.go#L54-L65). The only problem is that glog doesn't have debug/trace functions.

mrd0ll4r:
Alright, I wrote [github.com/mrd0ll4r/logger](https://github.com/mrd0ll4r/logger), fork of [github.com/calmh/logger](https://github.com/calmh/logger).

I incorporated lots of the ideas and useful things from glog and calhm/logger and came up with this:

    // Logger is an interface for leveled loggers.
    // All logging methods must be thread-safe, no guarantee is provided regarding the other methods.
    type Logger interface {
    	// SetLevel instructs the logger to log everything above and including the provided level
    	SetLevel(LogLevel)
[...]

PR #125, opened 2015-11-28
Moved from glog to logger

mrd0ll4r:
This somewhat contributes to #44.

Note that logger does not (yet?) support mapping some log events to special outputs (i.e. `stderr`), so now everything just logs to `stdout` by default.

### DIFF
--- a/chihaya.go
+++ b/chihaya.go
@@ -14,8 +14,7 @@ import (
 	"runtime"
 	"sync"
 	"syscall"
-
-	"github.com/golang/glog"
+	"log"
 
 	"github.com/chihaya/chihaya/api"
 	"github.com/chihaya/chihaya/config"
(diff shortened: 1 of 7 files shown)


mrd0ll4r:
Huh, I just noticed this introduces output to the tests - should that be fixed?

chihaya.go
```diff
@@ -26,16 +24,22 @@ import (
...
+	maxProcs        int
+	configPath      string
+	logLevel        int
+	logFileLocation bool
 )
 
 func init() {
 	flag.IntVar(&maxProcs, "maxprocs", runtime.NumCPU(), "maximum parallel threads")
 	flag.StringVar(&configPath, "config", "", "path to the configuration file")
+	flag.IntVar(&logLevel, "level", int(logger.LevelInfo), "level of logging, where 0=Everything, 1=Trace, 2=Debug, 3=Info, 4=Warnings, 5=Fatal")
```
jzelinskie: Can we make this a string and have them literally pass the name of the log levels?

jzelinskie:
Also, if we turn on log everything it would be nice to use the glog format, else just use the normal format.

mrd0ll4r:
with glog format I guess you mean logging file locations by default?

jzelinskie:
yep

jzelinskie:
We're frozen at 1.0 until our rewrite becomes master.

PR #128, opened 2016-01-04
remove all private tracker logic

jzelinskie:
There are no consumers of any of this logic nor is it complete. We're
better off without it in the meantime until we have a use case and a
more cohesive model for expressing it.


pushrax:
:ok:

PR #129, opened 2016-01-05
add support for jwt validation of infohashes

jzelinskie:
This change supports JWT on HTTP using the `jwt` query parameter. The JWT verifies the infohash as being approved by the `issuer`, which is discovered from the `JWKSetURI` which is a part of the JWT spec.


stats/stats.go
```diff
@@ -122,7 +122,7 @@ func New(cfg config.StatsConfig) *Stats {
 
 	if cfg.IncludeMem {
 		s.MemStatsWrapper = NewMemStatsWrapper(cfg.VerboseMem)
-		s.recordMemStats = time.NewTicker(cfg.MemUpdateInterval.Duration).C
+		s.recordMemStats = time.After(cfg.MemUpdateInterval.Duration)
```
pushrax: Should be `time.Tick`

tracker/jwt.go
```diff
@@ -0,0 +1,147 @@
...
+
+	if len(jwks.Keys) == 0 {
+		return errors.New("Failed to find any keys from JWK Set URI")
+	}
+
+	if jwks.Issuer == "" {
+		return errors.New("Failed to find any issuer from JWK Set URI")
+	}
+
+	ttl, _, _ := oidchttp.Cacheable(resp.Header)
```
pushrax: Do we really need to import this whole library just for one function?

tracker/jwt.go
```diff
@@ -0,0 +1,147 @@
...
+		return errors.New("Failed to verify JWT with all available verifiers")
+	}
+
+	claims, err := jwt.Claims()
+	if err != nil {
+		return err
+	}
+
+	if claimedIssuer, ok, err := claims.StringClaim("iss"); claimedIssuer != jwkSet.Issuer || err != nil || !ok {
+		return errors.New("Failed to validate JWT issuer claim")
```
pushrax: The first char in errors is typically lowercase right?

pushrax:
Pretty much LGTM. Not sure if you want this to be merged before the refactor we were discussing though.

PR #130, opened 2016-02-16
IP store, middlewares

mrd0ll4r:
- Added an IPStore 
- Removed unnecessary `name` parameter for Open(Client|Peer|IP)Store
- Made it so that (Client|Peer|IP)Stores are opened
- Added myself to the CONTRIBUTORS
- Added IP blacklist/whitelist middlewares

### DIFF
--- a/cmd/chihaya/main.go
+++ b/cmd/chihaya/main.go
@@ -15,6 +15,7 @@ import (
 	"github.com/chihaya/chihaya/server"
 	"github.com/chihaya/chihaya/tracker"
 
+	_ "github.com/chihaya/chihaya/server/store/middleware/ip"
 	_ "github.com/chihaya/chihaya/server/http"
 )
 
--- a/server/store/memory/ip_store.go
+++ b/server/store/memory/ip_store.go
@@ -1,0 +1,159 @@
+// Copyright 2016 The Chihaya Authors. All rights reserved.
+// Use of this source code is governed by the BSD 2-Clause license,
+// which can be found in the LICENSE file.
+
+package memory
+
+import (
+	"net"
+	"sync"
+
+	"github.com/mrd0ll4r/netmatch"
+
+	"github.com/chihaya/chihaya/server/store"
+)
+
+func init() {
+	store.RegisterIPStoreDriver("memory", &ipStoreDriver{})
+}
+
+type ipStoreDriver struct{}
+
+func (d *ipStoreDriver) New(cfg *store.Config) (store.IPStore, error) {
+	return &ipStore{
+		ips:      make(map[[16]byte]struct{}),
+		networks: netmatch.New(),
+	}, nil
+}
+
+// ipStore implements store.IPStore using an in-memory map of byte arrays and
+// a trie-like structure.
+type ipStore struct {
+	ips      map[[16]byte]struct{}
+	networks *netmatch.Trie
+	sync.RWMutex
+}
+
+var (
+	_            store.IPStore = &ipStore{}
+	v4InV6Prefix               = []byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xff, 0xff}
+)
+
+// key converts an IP address to a [16]byte.
+// The byte array can then be used as a key for a map, unlike net.IP, which is a
+// []byte.
+// If an IPv4 address is specified, it will be prefixed with
+// the net.v4InV6Prefix and thus becomes a valid IPv6 address.
+func key(ip net.IP) [16]byte {
+	var array [16]byte
+
+	if len(ip) == net.IPv4len {
+		copy(array[:], v4InV6Prefix)
+		copy(array[12:], ip)
+	} else {
+		copy(array[:], ip)
+	}
+	return array
+}
+
+func (s *ipStore) AddNetwork(network string) error {
+	s.Lock()
+	defer s.Unlock()
+
+	key, length, err := netmatch.ParseNetwork(network)
+	if err != nil {
+		return err
+	}
+
+	return s.networks.Add(key, length)
+}
+
+func (s *ipStore) AddIP(ip net.IP) error {
+	s.Lock()
+	defer s.Unlock()
+
+	s.ips[key(ip)] = struct{}{}
+
+	return nil
+}
+
+func (s *ipStore) HasIP(ip net.IP) (bool, error) {
+	s.RLock()
+	defer s.RUnlock()
+	key := key(ip)
+
+	_, ok := s.ips[key]
+	if ok {
+		return true, nil
+	}
+
+	match, err := s.networks.Match(key)
+	if err != nil {
+		return false, err
+	}
+
+	return match, nil
+}
+
+func (s *ipStore) HasAnyIP(ips []net.IP) (bool, error) {
+	s.RLock()
+	defer s.RUnlock()
+
+	for _, ip := range ips {
+		key := key(ip)
+		if _, ok := s.ips[key]; ok {
+			return true, nil
+		}
+
+		match, err := s.networks.Match(key)
+		if err != nil {
+			return false, err
+		}
+		if match {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func (s *ipStore) HasAllIPs(ips []net.IP) (bool, error) {
+	s.RLock()
+	defer s.RUnlock()
+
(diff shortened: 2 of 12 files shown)


AUTHORS
```diff
@@ -2,4 +2,4 @@
 
 Jimmy Zelinskie <jimmyzelinskie@gmail.com>
 Justin Li <jli@j-li.net>
-
+Leo Balduf <balduf@hm.edu>
```
jzelinskie: As discussed on IRC, let's remake the CONTRIBUTORS file.

middlewares/ip/README.md
```diff
@@ -0,0 +1,33 @@
+## IP Blacklisting/Whitelisting Middlewares
+
+This package provides the announce middlewares `ip_blacklist` and `ip_whitelist` for blacklisting or whitelisting 
+IP addresses for announces.
```
jzelinskie: For markdown, I like to put one sentence per line. It'll still be rendered into paragraphs, but it makes code changes more clear.

middlewares/ip/ip_blacklist.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func init() {
+	tracker.RegisterAnnounceMiddleware("ip_blacklist", blacklistAnnounceIP)
+}
+
+// ErrBlockedIP is returned by an announce middleware if any of the announcing IPs is disallowed
+var ErrBlockedIP = errors.NewMessage("disallowed IP address")
+
+// blacklistAnnounceIP provides a middleware that only allows IPs to announce that are not stored in an IPStore
```
jzelinskie: Go comments use 80 char line length.
Go code has no line length.
It's weird, but that's how the world works.

middlewares/ip/ip_blacklist.go
```diff
@@ -0,0 +1,48 @@
...
+func blacklistAnnounceIP(next tracker.AnnounceHandler) tracker.AnnounceHandler {
+	return func(cfg *config.TrackerConfig, req *chihaya.AnnounceRequest, resp *chihaya.AnnounceResponse) error {
+		var (
+			blacklisted bool
+			err         error
+		)
+
+		// we have to check explicitly if they are present, because someone could have added a <nil> net.IP to the store
```
jzelinskie: Go comments are also full sentences. With capitalization and punctuation.

middlewares/ip/ip_whitelist.go
```diff
@@ -0,0 +1,44 @@
...
+func init() {
+	tracker.RegisterAnnounceMiddleware("ip_whitelist", whitelistAnnounceIP)
+}
+
+// whitelistAnnounceIP provides a middleware that only allows IPs to announce that are stored in an IPStore
+func whitelistAnnounceIP(next tracker.AnnounceHandler) tracker.AnnounceHandler {
+	return func(cfg *config.TrackerConfig, req *chihaya.AnnounceRequest, resp *chihaya.AnnounceResponse) error {
+		var (
+			whitelisted bool
+			err         error
```
jzelinskie: you can save some lines if you change the signature of the function to return `(err error)` and then this block turns into just a `var whitelisted bool` or a `whitelisted := false`

server/store/memory/ip_store.go
```diff
@@ -0,0 +1,103 @@
...
+
+// toBytes converts an IP address to a [16]byte.
+// If an IPv4 address is specified, it will be prefixed with
+// the net.v4InV6Prefix and thus becomes a valid IPv6 address.
+func toBytes(ip net.IP) [16]byte {
+	var array [16]byte
+
+	if len(ip) == net.IPv4len {
+		copy(array[:], v4InV6Prefix)
+		copy(array[12:], ip)
```
jzelinskie: Why are you doing this instead of `ip.To6()`

server/store/memory/ip_store.go
```diff
@@ -0,0 +1,103 @@
...
+
+var (
+	_            store.IPStore = &ipStore{}
+	v4InV6Prefix               = []byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xff, 0xff}
+)
+
+// toBytes converts an IP address to a [16]byte.
+// If an IPv4 address is specified, it will be prefixed with
+// the net.v4InV6Prefix and thus becomes a valid IPv6 address.
+func toBytes(ip net.IP) [16]byte {
```
jzelinskie: Add something in the comment that they need to be an array so that it can be a key for the map.

server/store/memory/ip_store_test.go
```diff
@@ -0,0 +1,131 @@
+package memory
```
jzelinskie: copyright header

cmd/chihaya/main.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/chihaya/chihaya/server"
 	"github.com/chihaya/chihaya/tracker"
 
+	_ "github.com/chihaya/chihaya/middlewares/ip"
```
jzelinskie: I had thoughts about putting store-dependent middleware in a subpackage of the store i.e. `github.com/chihaya/chihaya/server/store/middleware/ip` the dependency on the store might appear more obvious then.

Thoughts?

middlewares/ip/README.md
```diff
@@ -0,0 +1,32 @@
+## IP Blacklisting/Whitelisting Middlewares
+
+This package provides the announce middlewares `ip_blacklist` and `ip_whitelist` for blacklisting or whitelisting IP addresses and networks for announces.
+
+### `ip_blacklist`
```
jzelinskie: All of our other YAML uses camelCase and not snake_case. YAML doesn't actually define anywhere which is the standard. If you think snake_case is better, open another PR that changes this across the board. But for now, let's put this in as camelCase.

middlewares/ip/ip_whitelist.go
```diff
@@ -0,0 +1,43 @@
...
+// whitelistAnnounceIP provides a middleware that only allows IPs to announce
+// that are stored in an IPStore.
+func whitelistAnnounceIP(next tracker.AnnounceHandler) tracker.AnnounceHandler {
+	return func(cfg *config.TrackerConfig, req *chihaya.AnnounceRequest, resp *chihaya.AnnounceResponse) (err error) {
+		var whitelisted bool
+
+		// We have to check explicitly if they are present, because someone
+		// could have added a <nil> net.IP to the store.
+		if req.IPv4 != nil && req.IPv6 != nil {
+			whitelisted, err = store.MustGetStore().HasAllIPs([]net.IP{req.IPv4, req.IPv6})
```
jzelinskie: just do one `storage := store.MustGetStore()` and then reuse it.

server/store/memory/ip_store.go
```diff
@@ -0,0 +1,159 @@
...
+
+package memory
+
+import (
+	"net"
+	"sync"
+
+	"github.com/chihaya/chihaya/server/store"
+
+	"github.com/mrd0ll4r/netmatch"
```
jzelinskie: third party imports go above internal ones

cmd/chihaya/main.go
mrd0ll4r: It kinda makes sense - then again it would be in a subpackage of `github.com/chihaya/chihaya/server`, but it's definitely no server. (One might even go so far as to say that `github.com/chihaya/chihaya/server/store/middlewares` could be an implementation of `github.com/chihaya/chihaya/server/store`, just like `memory` is)

I think I'll move it, just to keep `chihaya/middlewares` nice and clean for non-server-dependent middlewares.

middlewares/ip/ip_whitelist.go
mrd0ll4r: That's what I had in mind, but currently it shouldn't work:

In `cmd/chihaya/main.go`: We open the tracker first, i.e. the `whitelistAnnounceIP` function will be called before the store "server" is started (we start the pool after initializing the tracker, which makes sense) - if I try to get the store then, it'll fail

jzelinskie: This function just returns the inner-function, though. It doesn't get called until a request flows through the tracker. I change I suggested was to remove the verbosity of calling MustGetStore() over and over again, it shouldn't change the semantics of anything.

jzelinskie:
LGTM

PR #132, opened 2016-02-28
Client middlewares

mrd0ll4r:
Added client middlewares, fixed ClientStore to use `chihaya.PeerID`s for lookup


jzelinskie:
LGTM! Thanks!

PR #134, opened 2016-02-28
store: added StringStore

mrd0ll4r:
One step at a time, just adding the StringStore for now.

I thought about adding/checking/removing not the complete string, but rather a prefix plus the data - this would allow us to have smaller locks, if that makes sense...

### DIFF
--- a/server/store/string_store.go
+++ b/server/store/string_store.go
@@ -1,0 +1,49 @@
+// Copyright 2016 The Chihaya Authors. All rights reserved.
+// Use of this source code is governed by the BSD 2-Clause license,
+// which can be found in the LICENSE file.
+
+package store
+
+import "fmt"
+
+// PrefixInfohash is the prefix to be used for infohashes.
+const PrefixInfohash = "ih-"
+
+var stringStoreDrivers = make(map[string]StringStoreDriver)
+
+// StringStore represents an interface for manipulating strings.
+type StringStore interface {
+	PutString(s string) error
+	HasString(s string) (bool, error)
+	RemoveString(s string) error
+}
+
+// StringStoreDriver represents an interface for creating a handle to the
+// storage of swarms.
+type StringStoreDriver interface {
+	New(*Config) (StringStore, error)
+}
+
+// RegisterStringStoreDriver makes a driver available by the provided name.
+//
+// If this function is called twice with the same name or if the driver is nil,
+// it panics.
+func RegisterStringStoreDriver(name string, driver StringStoreDriver) {
+	if driver == nil {
+		panic("store: could not register nil StringStoreDriver")
+	}
+	if _, dup := stringStoreDrivers[name]; dup {
+		panic("store: could not register duplicate StringStoreDriver: " + name)
+	}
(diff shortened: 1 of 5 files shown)


server/store/string_store.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2016 The Chihaya Authors. All rights reserved.
+// Use of this source code is governed by the BSD 2-Clause license,
+// which can be found in the LICENSE file.
+
+package store
+
+import "fmt"
+
+// PrefixInfohash is the prefix to be used for infohashes.
+const PrefixInfohash = "ih-"
```
jzelinskie: I think this should be defined in the middleware.

jzelinskie:
>I thought about adding/checking/removing not the complete string, but rather a prefix plus the data - this would allow us to have smaller locks, if that makes sense...

I'm afraid I don't follow.

server/store/string_store.go
mrd0ll4r: The idea was to include the most common prefixes here, to avoid collisions. Store dependent middleware has to use the store anyway, so all store dependent middlewares have access to this.

mrd0ll4r:
Instead of
```go
func PutString(s string) error
...
```
we could use
```go
func PutString(prefix,s string) error
...
```

This would allow for a synchronization similar to the sharding we use for peers.

jzelinskie:
I don't want to shard on just the prefix, though. If we have tons of infohashes, we want to be able to continue sharding them beyond just one shard for the `ih-` prefix.

jzelinskie:
I rebased on master which means this PR also needs to be rebased.

I'm thinking of cutting a "1.0.0" tag where master is currently and merging the middleware branch into master.

jzelinskie:
I rebased on master which means this PR also needs to be rebased.

I'm thinking of cutting a "1.0.0" tag where master is currently and merging the middleware branch into master and start calling it "2.0.x". Thoughts @pushrax?

mrd0ll4r:
About the sharding: I agree with you and seperating the prefix from the actual string helps with this:
If we had just one parameter, we'd have to split it into prefix and string and then shard that on every lookup, put and remove/delete.

About merging into master: I personally feel it's too early, we should at least get the most basic swarm/peer middleware done, so we actually have a working tracker

jzelinskie:
>If we had just one parameter, we'd have to split it into prefix and string and then shard that on every lookup, put and remove/delete.

That's not necessarily true. We could just hash the whole thing and ignore the prefix from the shard perspective and simply use it for the separation of namespaces. However, you're suggesting something like `map[prefix]map[string]struct{}`?

mrd0ll4r:
Yep, that's what I had in mind.The whole issue we're discussing here is pretty implementation-focused, I already had the performance of the implementation in mind when suggesting to have to parameters (to avoid splitting).

Hashing and sharding based on that is an option, true. I'll leave the signature as it is right now and add a test to make coveralls happy

jzelinskie:
LGTM. Thanks! What are your plans for follow up PRs?

PR #135, opened 2016-03-02
Hotfix configparse

mrd0ll4r:
This
 - fixes config parsing for both the store and the http server
 - fixes synchronization when shutting down the pool
 - adds a log line indicating the HTTP server has been shut down cleanly
 - adds dependencies to cmd/chihaya, so that more shipped modules are built


jzelinskie:
LGTM

PR #138, opened 2016-03-11
Config normalize

mrd0ll4r:
- Made it so that all embedded driver/config constructs use the same format (`Name` and `Config`)
- Fixed some error messages/code formatting
- Added easier to debug error messages in case a driver for the store is missing (a driver for what store exactly?)


jzelinskie:
LGTM!

PR #139, opened 2016-03-24
tracker: added a way for middleware constructors to fail

mrd0ll4r:
If for example the config is invalid, a *MiddlewareConstructor should be able to fail without panicking


jzelinskie:
LGTM

PR #140, opened 2016-03-24
added goveralls workaround

mrd0ll4r:
The issue seems to be [known](https://github.com/mattn/goveralls/issues/68) and this seems to be the way to fix it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,17 @@ go:
 sudo: false
 
 install:
+  - go get github.com/onsi/ginkgo/ginkgo
   - go get github.com/axw/gocov/gocov
   - go get github.com/mattn/goveralls
+  - go get golang.org/x/tools/cmd/cover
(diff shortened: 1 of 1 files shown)


jzelinskie:
Doesn't look like it worked. Can you try force pushing to see if it was just an ephemeral issue or if this is still broken?

mrd0ll4r:
nope, doesn't look like it works

mrd0ll4r:
Ah, I think travis is hiding the coveralls token, as described in the [travis docs about environment variables and pull requests](https://docs.travis-ci.com/user/environment-variables)

jzelinskie:
Closing in favor of #144. Thanks for the PR, but it looks like the tools for integrating with coveralls are just too much of a hack. See https://github.com/mattn/goveralls/pull/64 and https://github.com/mattn/goveralls/issues/68#issuecomment-198573259

PR #141, opened 2016-03-29
Godep: replace by glide/vendor

### LINKED ISSUE #85: Move from godep to experimental Go 1.5 vendor

joshdekock:
This fixes #85, and makes chihaya more inline with how dependencies are being handled as of Go 1.5. If there's anything I missed, then please let me know


joshdekock:
I seem to have missed changing the `.travis.yml` and forgot to remove a chihaya binary. Fixing that currently.

joshdekock:
Moving fix to the `middleware-refractor` branch instead.

jzelinskie:
I'll just note this here: the way we previously decided to use Godep was that we don't actually vendor any of the code. Rather, we just have the JSON file that has the SHAs locked and use `godep restore` to get all of the dependencies into the `GOPATH`. This means not having to keep a whole bunch of vendored code in our project.

PR #143, opened 2016-03-29
http: stopped throwing error on invalid event

joshdekock:
Event is an optional key, and if it's invalid or non-existent then it can just be assumed as 'none'.

### DIFF
--- a/server/http/request.go
+++ b/server/http/request.go
@@ -23,8 +23,10 @@ func announceRequest(r *http.Request, cfg *httpConfig) (*chihaya.AnnounceRequest
 	request := &chihaya.AnnounceRequest{Params: q}
 
 	eventStr, err := q.String("event")
-	if err != nil {
-		eventStr = "none"
+	if err == query.ErrKeyNotFound {
+		eventStr = ""
+	} else if err != nil {
+		return nil, tracker.ClientError("failed to parse parameter: event")
 	}
 	request.Event, err = event.New(eventStr)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


server/http/request.go
```diff
@@ -24,7 +24,7 @@ func announceRequest(r *http.Request, cfg *httpConfig) (*chihaya.AnnounceRequest
 
 	eventStr, err := q.String("event")
 	if err != nil {
-		return nil, tracker.ClientError("failed to parse parameter: event")
+		eventStr = "none"
```
mrd0ll4r: maybe we should check if `err == query.ErrKeyNotFound`? It's the only possible error returned right now, but that could change in the future.

@jzelinskie ?

jzelinskie:
We actually handle empty string to be the "none" event see [this line](https://github.com/chihaya/chihaya/blob/0652aefcb545737ce4e031fa5492a88ba8457a78/pkg/event/event.go#L47). We should rewrite this error block to be `if err == query.ErrKeyNotFound { event = "" }`

joshdekock:
and if it isn't `query.ErrKeyNotFound`? Just throw an error as before?

jzelinskie:
Yep. At the moment, `ErrKeyNotFound` is the only possible error returned from the string function: https://github.com/chihaya/chihaya/blob/6461e28b0f43941e766e60f3578f6f77447d84da/server/http/query/query.go#L99

jzelinskie:
LGTM. If you want to add yourself to the contributors file, you can and then I'll merge the PR!

joshdekock:
@jzelinskie I've added myself to the contributors file

PR #147, opened 2016-03-31
Add peer/seed info helper functions, and tests

joshdekock:


joshdekock:
oops, I forgot to rebase

server/store/memory/peer_store_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/chihaya/chihaya/server/store"
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	hash1 = chihaya.InfoHash("11111111111111111111")
+	hash2 = chihaya.InfoHash("22222222222222222222")
+	hash3 = chihaya.InfoHash("33333333333333333333")
+
+	peers = []chihaya.Peer{
```
jzelinskie: I like to make test data like this the most readable as possible and then have a function that creates the right type.

For example this could be

```go
var []struct{
  seeder bool
  peerID string
  ip string
  port int32
}{
  {true, "-AZ3034-6wfG2wk6wWLc", "127.0.0.1", 5000},
}
```

You can find some example PeerIDs in [this file](https://github.com/chihaya/chihaya/blob/e2da9cdc3ebf5eec300773f9aaa894123ce3ba6a/pkg/clientid/client_id_test.go)

server/store/memory/peer_store_test.go
```diff
@@ -0,0 +1,93 @@
...
+	var num uint32
+
+	s, err := d.New(&store.DriverConfig{
+		"memory",
+		struct{ Shards int }{1},
+	})
+	assert.Nil(t, err)
+	assert.NotNil(t, s)
+
+	// add all peers, odd indexes are seeders, and even are leechers
```
jzelinskie: This is very error-prone. You should encode this into your test data as in my previous comment.

server/store/peer_store.go
```diff
@@ -24,6 +24,12 @@ type PeerStore interface {
 	GraduateLeecher(infoHash chihaya.InfoHash, p chihaya.Peer) error
 	AnnouncePeers(infoHash chihaya.InfoHash, seeder bool, numWant int) (peers, peers6 []chihaya.Peer, err error)
 	CollectGarbage(cutoff time.Time) error
+
+	GetSeeders(infoHash chihaya.InfoHash) (peers, peers6 []chihaya.Peer, err error)
+	GetLeechers(infoHash chihaya.InfoHash) (peers, peers6 []chihaya.Peer, err error)
```
jzelinskie: This seems logical for now, but for example Facebook's subnet middleware is going to want to know the client's IP address to pick out the seeders. Maybe we should change all of these methods to use `AnnounceRequest` or `ScrapeRequest`.

Thoughts? @pushrax @mrd0ll4r

jzelinskie:
Congrats on the first PR!

joshdekock:
@jzelinskie I've made the suggested edits

server/store/memory/peer_store.go
```diff
@@ -88,16 +90,19 @@ func leechersKey(infoHash chihaya.InfoHash) string {
 
 func (s *peerStore) PutSeeder(infoHash chihaya.InfoHash, p chihaya.Peer) error {
 	key := seedersKey(infoHash)
+	i := s.shardIndex(infoHash)
+	if s.shards[i] == nil {
```
mrd0ll4r: Better initialize all shards in `New` already, easier to read and saves the `nil` checks on every call

server/store/memory/peer_store.go
```diff
@@ -127,16 +132,20 @@ func (s *peerStore) DeleteSeeder(infoHash chihaya.InfoHash, p chihaya.Peer) erro
 
 func (s *peerStore) PutLeecher(infoHash chihaya.InfoHash, p chihaya.Peer) error {
 	key := leechersKey(infoHash)
+	i := s.shardIndex(infoHash)
 
-	shard := s.shards[s.shardIndex(infoHash)]
-	shard.Lock()
-	defer shard.Unlock()
+	if s.shards[i] == nil {
```
mrd0ll4r: same as above

server/store/memory/peer_store.go
jzelinskie: I think I initially wrote it like this to attempt to save memory, but in practice the allocations probably just slow down the whole program.

server/store/memory/peer_store_test.go
```diff
@@ -0,0 +1,116 @@
...
+	for _, v := range peers {
+		if v.ID == peer.ID && v.Port == peer.Port {
+			return true
+		}
+	}
+	return false
+}
+
+func TestPeerStore(t *testing.T) {
+	var d = &peerStoreDriver{}
```
mrd0ll4r: make this a `var(...)` block for style points

server/store/memory/peer_store.go
```diff
@@ -280,3 +289,51 @@ func (s *peerStore) AnnouncePeers(infoHash chihaya.InfoHash, seeder bool, numWan
...
+		} else {
+			peers = append(peers, p.Peer)
+		}
+	}
+	return
+}
+
+func (s *peerStore) NumSeeders(infoHash chihaya.InfoHash) (uint32, error) {
+	seeders, seeders6, err := s.GetSeeders(infoHash)
+	seeders = append(seeders, seeders6...)
```
mrd0ll4r: save memory allocations by just adding their length

server/store/peer_store.go
mrd0ll4r: Facebook's subnet middleware will have the request though, or did I get you wrong?

server/store/peer_store.go
```diff
@@ -24,6 +24,12 @@ type PeerStore interface {
...
+
+	NumSeeders(infoHash chihaya.InfoHash) (uint32, error)
+	NumLeechers(infoHash chihaya.InfoHash) (uint32, error)
```
mrd0ll4r: Why do these return `uint32`s?

joshdekock: It was to preserve memory, but I think they could actually be `uint16`

server/store/memory/peer_store.go
joshdekock: Don't know how I missed that

server/store/memory/peer_store_test.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	config = store.DriverConfig{
+		"memory",
+		unmarshalledConfig,
+	}
+)
+
+func peerInSlice(peer chihaya.Peer, peers []chihaya.Peer) bool {
+	for _, v := range peers {
+		if v.ID == peer.ID && v.Port == peer.Port {
```
mrd0ll4r: You could implement an `Equal` or `CompareTo` (or ...) method on `chihaya.Peer`, sounds like something we could reuse.

server/store/peer_store.go
```diff
@@ -24,6 +24,12 @@ type PeerStore interface {
 	GraduateLeecher(infoHash chihaya.InfoHash, p chihaya.Peer) error
 	AnnouncePeers(infoHash chihaya.InfoHash, seeder bool, numWant int) (peers, peers6 []chihaya.Peer, err error)
 	CollectGarbage(cutoff time.Time) error
+
+	GetSeeders(infoHash chihaya.InfoHash) (peers, peers6 []chihaya.Peer, err error)
```
mrd0ll4r: godoc all of them

server/store/memory/peer_store.go
mrd0ll4r: You can even save the call to `GetSeeders` :)

server/store/peer_store.go
jzelinskie: Sure, but they should also be able to reuse this store interface and just have an alternative memory implementation.

server/store/peer_store.go
mrd0ll4r: I had `int` in mind because it's so much easier to handle without having to convert back and forth between `uint` and `int`, plus the length of a map will always fit into an int afaik

server/store/memory/peer_store.go
jzelinskie: We really should be internally tracking these lengths rather than trying to add them up each scrape request.

Also the `Get(Seeders|Leechers)` methods should take a "numWant int" argument so that they know when to stop.

mrd0ll4r: Implement `NumSeeders` like this:

```go
skey := seedersKey(infoHash)

shard := s.shards[s.shardIndex(infoHash)]
shard.RLock()
defer shard.RUnlock()

return len(shard.peers[skey])
```

Although I don't know how performant `len` calls on maps are - I have a repo about microbenchmarks, I should add that.

I thought the idea behind `Get(Seeder|Leechers)` was to get all seeders/leechers so that the middleware can figure out which ones it wants?

jzelinskie: All uses of the `len` builtin in Go are O(1)

chihaya_test.go
```diff
@@ -0,0 +1,44 @@
...
+		{"-AZ3034-6wfG2wk6wWLc", "250.183.81.177", 5720},
+		{"-BS5820-oy4La2MWGEFj", "fd45:7856:3dae::48", 2878},
+		{"-TR0960-6ep6svaa61r4", "fd45:7856:3dae::48", 2878},
+		{"-BS5820-oy4La2MWGEFj", "fd0a:29a8:8445::38", 2878},
+		{"-BS5820-oy4La2MWGEFj", "fd45:7856:3dae::48", 8999},
+	}
+	builtPeers []*Peer
+)
+
+func TestChihaya(t *testing.T) {
```
mrd0ll4r: The name is kind of  misleading

server/store/memory/peer_store_test.go
```diff
@@ -0,0 +1,117 @@
...
+		} else {
+			err = s.PutLeecher(hash, peer)
+		}
+		assert.Nil(t, err)
+	}
+
+	leechers1, leechers61, err := s.GetLeechers(hash)
+	assert.NotEmpty(t, leechers1)
+	assert.NotEmpty(t, leechers61)
+	assert.Nil(t, err)
```
mrd0ll4r: the order is wrong: check the error first, then the results

server/store/memory/peer_store_test.go
```diff
@@ -0,0 +1,117 @@
...
+	leechers1 = append(leechers1, leechers61...)
+	num = s.NumLeechers(hash)
+	assert.Nil(t, err)
```
mrd0ll4r: `err` did not change since the last check

server/store/peer_store.go
```diff
@@ -15,15 +15,39 @@ var peerStoreDrivers = make(map[string]PeerStoreDriver)
 
 // PeerStore represents an interface for manipulating peers.
 type PeerStore interface {
+	// PutSeeder adds a seeder for the infoHash to the PeerStore
 	PutSeeder(infoHash chihaya.InfoHash, p chihaya.Peer) error
+	// PutSeeder removes a seeder for the infoHash from the PeerStore
```
mrd0ll4r: Typo, should be `DeleteSeeder` - let golint check this stuff for you :)

server/store/peer_store.go
```diff
@@ -15,15 +15,39 @@ var peerStoreDrivers = make(map[string]PeerStoreDriver)
...
 	DeleteSeeder(infoHash chihaya.InfoHash, p chihaya.Peer) error
 
+	// PutLeecher adds a seeder for the infoHash to the PeerStore
```
mrd0ll4r: adds a leecher, afaik, same with `DeleteLeecher`

server/store/memory/peer_store.go
```diff
@@ -127,16 +132,15 @@ func (s *peerStore) DeleteSeeder(infoHash chihaya.InfoHash, p chihaya.Peer) erro
 
 func (s *peerStore) PutLeecher(infoHash chihaya.InfoHash, p chihaya.Peer) error {
 	key := leechersKey(infoHash)
+	i := s.shardIndex(infoHash)
```
mrd0ll4r: you can go back to using `shard := s.shards[s.shardIndex(infoHash)]`, like you did for `GetSeeders`

chihaya_test.go
```diff
@@ -0,0 +1,44 @@
...
+		{"-TR0960-6ep6svaa61r4", "fd45:7856:3dae::48", 2878},
+		{"-BS5820-oy4La2MWGEFj", "fd0a:29a8:8445::38", 2878},
+		{"-BS5820-oy4La2MWGEFj", "fd45:7856:3dae::48", 8999},
+	}
+	builtPeers []*Peer
+)
+
+func TestPeer(t *testing.T) {
+	// build peers from test data
+	for i := 0; i < len(peers); i++ {
```
jzelinskie: `for p := range peers { ... }`

server/store/memory/peer_store.go
```diff
@@ -280,3 +284,59 @@ func (s *peerStore) AnnouncePeers(infoHash chihaya.InfoHash, seeder bool, numWan
 
 	return
 }
+
+func (s *peerStore) GetSeeders(infoHash chihaya.InfoHash) (peers, peers6 []chihaya.Peer, err error) {
```
mrd0ll4r: I still feel like these return all seeders, not just numWant?

server/store/peer_store.go
```diff
@@ -15,15 +15,39 @@ var peerStoreDrivers = make(map[string]PeerStoreDriver)
...
+	// seeders, whichever is smaller. If the available seeders is less than numWant
+	// then peers are returned until numWant or they run out.
 	AnnouncePeers(infoHash chihaya.InfoHash, seeder bool, numWant int) (peers, peers6 []chihaya.Peer, err error)
+	// CollectGarbage deletes peers from the peerStore which are older than the cutoff time
 	CollectGarbage(cutoff time.Time) error
+
+	// GetSeeders gets all the seeders for a particular infoHash
+	GetSeeders(infoHash chihaya.InfoHash) (peers, peers6 []chihaya.Peer, err error)
+	// GetLeechers gets all the leechers for a particular infoHash
+	GetLeechers(infoHash chihaya.InfoHash) (peers, peers6 []chihaya.Peer, err error)
```
jzelinskie: The only reason we have `GetSeeders` and `GetLeechers` is for the API, right? `AnnouncePeers` should handle it for the tracker aspect. I wonder if we could instead be clever and make it so that `AnnouncePeers` with `-1` for numWant returns all. Then we could just call it twice with seeder as `true` and `false` and then de-duplicate what we have. It makes the API slightly less efficient, but lessens the burden on implementing drivers. The API shouldn't be used nearly as much as the Tracker itself either.

server/store/peer_store.go
```diff
@@ -15,15 +15,39 @@ var peerStoreDrivers = make(map[string]PeerStoreDriver)
 
 // PeerStore represents an interface for manipulating peers.
 type PeerStore interface {
+	// PutSeeder adds a seeder for the infoHash to the PeerStore
```
jzelinskie: All the comments here need to end with periods.

Pedantic, but necessary.

server/store/memory/peer_store_test.go
```diff
@@ -0,0 +1,116 @@
...
+		ret  bool
+		num  int
+		peer chihaya.Peer
+	)
+	s, err := d.New(&config)
+	assert.Nil(t, err)
+	assert.NotNil(t, s)
+
+	for _, p := range peers {
+		// construct chihaya.Peer from test data
```
jzelinskie: Again, "sentence-like" comments.

server/store/memory/peer_store_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+func TestPeerStore(t *testing.T) {
+	var (
+		d    = &peerStoreDriver{}
+		err  error
+		ret  bool
+		num  int
+		peer chihaya.Peer
+	)
+	s, err := d.New(&config)
```
jzelinskie: No need to have `var err error` if you allocate it here.

Go style isn't like C or JS where you forward declare everything. Try and avoid big `var` blocks, lots of the time they aren't necessary. Try and declare variables closer to where they are used.

server/store/memory/peer_store.go
jzelinskie: Yep, I was confused. I forgot `AnnouncePeers` even existed.

server/store/memory/peer_store_test.go
```diff
@@ -0,0 +1,116 @@
...
+func peerInSlice(peer chihaya.Peer, peers []chihaya.Peer) bool {
+	for _, v := range peers {
+		if v.Equal(&peer) {
+			return true
+		}
+	}
+	return false
+}
+
+func TestPeerStore(t *testing.T) {
```
jzelinskie: Can this test either be named better or broken into smaller tests that test one thing?

joshdekock:
I have no idea how to test AnnouncePeers, maybe someone else could do that?

chihaya_test.go
```diff
@@ -0,0 +1,44 @@
...
+		ip     string
+		port   uint16
+	}{
+		{"-AZ3034-6wfG2wk6wWLc", "250.183.81.177", 5720},
+		{"-BS5820-oy4La2MWGEFj", "fd45:7856:3dae::48", 2878},
+		{"-TR0960-6ep6svaa61r4", "fd45:7856:3dae::48", 2878},
+		{"-BS5820-oy4La2MWGEFj", "fd0a:29a8:8445::38", 2878},
+		{"-BS5820-oy4La2MWGEFj", "fd45:7856:3dae::48", 8999},
+	}
+	builtPeers []*Peer
```
jzelinskie: pull this var into the test function

server/store/memory/peer_store.go
```diff
@@ -27,9 +27,13 @@ func (d *peerStoreDriver) New(storecfg *store.DriverConfig) (store.PeerStore, er
 	if err != nil {
 		return nil, err
 	}
-
+	shards := make([]*peerShard, cfg.Shards)
```
jzelinskie: keep the linebreak above this

server/store/memory/peer_store.go
```diff
@@ -48,7 +52,9 @@ func newPeerStoreConfig(storecfg *store.DriverConfig) (*peerStoreConfig, error)
 	if err != nil {
 		return nil, err
 	}
-
+	if cfg.Shards < 1 {
```
jzelinskie: linebreak above this

server/store/memory/peer_store_test.go
```diff
@@ -0,0 +1,166 @@
...
+	// Check that there are 7 seeders, and 3 leechers after graduating a
+	// leecher to a seeder.
+	assert.Equal(t, 7, s.NumSeeders(hash))
+	assert.Equal(t, 3, s.NumLeechers(hash))
+
+	peers1, peers61, err := s.AnnouncePeers(hash, true, 5)
+	assert.Nil(t, err)
+	assert.NotNil(t, peers1)
+	assert.NotNil(t, peers61)
+	// TODO: Properly test AnnouncePeers.
```
jzelinskie: remove this line and make a GitHub issue

server/store/peer_store.go
```diff
@@ -15,15 +15,42 @@ var peerStoreDrivers = make(map[string]PeerStoreDriver)
...
 
+	// PutLeecher adds a leecher for the infoHash to the PeerStore.
 	PutLeecher(infoHash chihaya.InfoHash, p chihaya.Peer) error
+	// DeleteLeecher removes a leecher for the infoHash from the PeerStore.
 	DeleteLeecher(infoHash chihaya.InfoHash, p chihaya.Peer) error
 
+	// GraduateLeecher promotes a peer from a leecher to a seeder for the
+	// infoHash within the PeerStore.
 	GraduateLeecher(infoHash chihaya.InfoHash, p chihaya.Peer) error
+	// AnnouncePeers returns a list of both ipv4, and ipv6 peers for an
```
jzelinskie: capitalize `IPv4` and `IPv6` correctly

chihaya_test.go
```diff
@@ -0,0 +1,43 @@
...
+			port   uint16
+		}{
+			{"-AZ3034-6wfG2wk6wWLc", "250.183.81.177", 5720},
+			{"-BS5820-oy4La2MWGEFj", "fd45:7856:3dae::48", 2878},
+			{"-TR0960-6ep6svaa61r4", "fd45:7856:3dae::48", 2878},
+			{"-BS5820-oy4La2MWGEFj", "fd0a:29a8:8445::38", 2878},
+			{"-BS5820-oy4La2MWGEFj", "fd45:7856:3dae::48", 8999},
+		}
+		builtPeers []*Peer
+	)
```
jzelinskie: I think we had a miscommunication.

I just wanted:

```go
var builtPeers []*Peers
for i := 0; i < len(peers); i ++ {
  ...
}
```

You can leave the tests in the top level scope.

server/store/memory/peer_store_test.go
```diff
@@ -0,0 +1,162 @@
...
+	for _, p := range peers {
+		// Construct chihaya.Peer from test data.
+		peer = chihaya.Peer{
+			chihaya.PeerID(p.peerID),
+			net.ParseIP(p.ip),
+			p.port,
+		}
+
+		if p.seeder {
+			ret = peerInSlice(peer, seeders)
```
jzelinskie: You can get rid of `ret` (terrible name btw) all together if you change these lines to be `assert.True(t, peerInSlice(peer, seeders)`

chihaya_test.go
```diff
@@ -0,0 +1,44 @@
...
+		{"-BS5820-oy4La2MWGEFj", "fd45:7856:3dae::48", 2878},
+		{"-TR0960-6ep6svaa61r4", "fd45:7856:3dae::48", 2878},
+		{"-BS5820-oy4La2MWGEFj", "fd0a:29a8:8445::38", 2878},
+		{"-BS5820-oy4La2MWGEFj", "fd45:7856:3dae::48", 8999},
+	}
+)
+
+func TestPeerEquality(t *testing.T) {
+	var builtPeers []*Peer
+	// Build peers from test data.
```
jzelinskie: move the comment above the var

chihaya_test.go
```diff
@@ -0,0 +1,44 @@
...
+	for i := 0; i < len(peers); i++ {
+		builtPeers = append(builtPeers, &Peer{
+			ID:   PeerID(peers[i].peerID),
+			IP:   net.ParseIP(peers[i].ip),
+			Port: peers[i].port,
+		})
+	}
```
jzelinskie: newline after this for loop

chihaya_test.go
```diff
@@ -0,0 +1,44 @@
...
+		{"-TR0960-6ep6svaa61r4", "fd45:7856:3dae::48", 2878},
+		{"-BS5820-oy4La2MWGEFj", "fd0a:29a8:8445::38", 2878},
+		{"-BS5820-oy4La2MWGEFj", "fd45:7856:3dae::48", 8999},
+	}
+)
+
+func TestPeerEquality(t *testing.T) {
+	var builtPeers []*Peer
+	// Build peers from test data.
+	for i := 0; i < len(peers); i++ {
```
jzelinskie: use `range` rather than a counter

server/store/memory/peer_store_test.go
```diff
@@ -0,0 +1,162 @@
...
+		ret  bool
+		peer chihaya.Peer
+	)
+	s, err := d.New(&config)
+	assert.Nil(t, err)
+	assert.NotNil(t, s)
+
+	for _, p := range peers {
+		// Construct chihaya.Peer from test data.
+		peer = chihaya.Peer{
```
jzelinskie: make this `:=` and you can get rid of the peer in the var block

jzelinskie:
LGTM

Issue #145, opened 2016-03-30
Potential Middleware


jzelinskie:
- [x] Client Whitelist/Blacklist
- [x] IP Whitelist/Blacklist
- [x] Infohash Whitelist/Blacklist
- [ ] Preferred Subnets
- [ ] JWT Verification
- [x] Basic Response

mrd0ll4r:
IP Whitelist/Blacklist only needs some tests, Infohash Whitelist/Blacklist has a PR open

I'll implement:

- [ ] Deniability (insert random peers to a response)
- [ ] Ratelimit

joshdekock:
- [ ] torrent caching (like [torrage](http://torrage.info/) but done automatically) 

_Although I'm not sure if this should be a server instead, as it will need to be able to seed a metafile asynchronously to the middleware stack (not that this couldn't be achieved with go statements)_

Also, just as a heads up, I'm implementing Basic Response currently

jzelinskie:
I don't understand what you mean by torrent caching. If I understand it correctly, it's just a lookup for piece data given an infohash. The BitTorrent DHT already does this.

If I was hypothetically trying to download something and I only had the infohash, I'd just open it with a client using a magnet URI and use the DHT to get metadata and then I just add _all_ the public trackers I know about to try and find peers.

joshdekock:
Yeah, I'm not sure I thought about it properly

jzelinskie:
Going to close this since there isn't much discussion going on, the topic is super general, and most of these have been implemented.

PR #152, opened 2016-04-02
middleware: added varinterval

mrd0ll4r:
Contributes to #145


jzelinskie:
Maybe we should have a section for the middleware which is a "use case" or "why/when would I use this"?

I suppose this one is for avoiding tons of peers all hitting the tracker at the same time and spiking traffic?

mrd0ll4r:
Yup, will add to the README.

jzelinskie:
LGTM

PR #157, opened 2016-04-04
more general interfaces

### LINKED ISSUE #156: interfaces
[interfacer](https://github.com/mvdan/interfacer) spits out this:

```
pkg/bencode/decoder.go:116:21: r can be io.ByteScanner
server/http/request.go:109:18: q can be chihaya.Params
server/http/writer.go:27:28: w can be io.Writer
server/http/writer.go:71:26: w can be io.Writer
```

Just FYI, dunno if we should actually change them.

jzelinskie:
Fixes #156.

We could later expand this to make all the current methods in `http/request.go` use `chihaya.Params` and wrap them with a method that specifically uses the `query` package.

### DIFF
--- a/pkg/bencode/decoder.go
+++ b/pkg/bencode/decoder.go
@@ -113,7 +113,7 @@ func unmarshal(r *bufio.Reader) (interface{}, error) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pushrax:
Sure :+1:

PR #159, opened 2016-04-04
Make client middleware use StringStore and remove ClientStore

### LINKED ISSUE #158: Replace Client Store usage with String Store
No point keeping around Client Store, as the string store was intended to be a generalization of its storage.

We should remove it from the `server/store` package and update the middleware to use string store.

joshdekock:
Fixes #158


mrd0ll4r:
Use just the client part of the peer ID and rename the prefix accordingly

joshdekock:
e.g. `PrefixPeerID` -> `PrefixClientPeerID` ?

mrd0ll4r:
I'd call it `PrefixClientID` - have a look at the pkg/clientid package for extracting that from a peer ID

server/store/string_store.go
```diff
@@ -9,6 +9,9 @@ import "fmt"
 // PrefixInfohash is the prefix to be used for infohashes.
 const PrefixInfohash = "ih-"
 
+// PrefixPeerID is the prefix to be used for client peer IDs.
+const PrefixPeerID = "pi-"
```
mrd0ll4r: I'd call this `PrefixClientID`. And I don't know about the value of the prefix, could just be `c-` if you ask me, I chose `ih-` for infohashes because it kinda made sense, but afaik we don't have rules for the prefixes.
@jzelinskie  ?

jzelinskie: TBH, I still don't like having any of these prefixes in the string store package. I think it should be the job of the middleware to make sure their prefixes don't collide with anyone else's.

server/store/middleware/client/whitelist.go
```diff
@@ -14,16 +15,22 @@ func init() {
 	tracker.RegisterAnnounceMiddleware("client_whitelist", whitelistAnnounceClient)
 }
 
+// PrefixPeerID is the prefix to be used for client peer IDs.
+const PrefixPeerID = "pi-"
```
mrd0ll4r: `PrefixClient(ID)` ?

server/store/middleware/infohash/config.go
```diff
@@ -12,6 +12,9 @@ import (
 	"github.com/chihaya/chihaya"
 )
 
+// PrefixInfohash is the prefix to be used for infohashes.
+const PrefixInfohash = "ih-"
```
mrd0ll4r: move this into either the blacklist or whitelist file

server/store/middleware/client/whitelist.go
```diff
@@ -14,16 +15,22 @@ func init() {
 	tracker.RegisterAnnounceMiddleware("client_whitelist", whitelistAnnounceClient)
 }
 
+// PrefixClient is the prefix to be used for client peer IDs.
+const PrefixClient = "pi-"
```
mrd0ll4r: I'd change this to be `c-` and the godoc to use client IDs instead of client peer IDs.

jzelinskie:
LGTM

PR #161, opened 2016-04-05
rename seeder to seed & leecher to leech

joshdekock:
Switch to use more conventional terminology when talking about downloaders and uploaders in regards to bittorrent, this was discussed briefly with @jzelinskie.

### DIFF
--- a/server/store/memory/peer_store.go
+++ b/server/store/memory/peer_store.go
@@ -86,16 +86,16 @@ func peerKey(p chihaya.Peer) string {
 	return string(p.IP) + string(p.ID)
 }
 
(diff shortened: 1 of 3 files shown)


mrd0ll4r:
binary in the commit

mrd0ll4r:
Tbh I think leecher is the correct term, see [Wikipedia](https://en.m.wikipedia.org/wiki/Glossary_of_BitTorrent_terms#Leech) and [the other wiki](https://wiki.theory.org/BitTorrentSpecification#Tracker_Response) (complete and incomplete). The theory one actually uses seeder too.

joshdekock:
Hmm. I was sure it was leech/seed, but nevermind.

PR #162, opened 2016-04-06
Lint, vet, Start/Stop

### LINKED ISSUE #46: Lint ALL the things
There should be no file that doesn't fully pass [golint](https://github.com/golang/lint).

This mostly just means adding lots of doc comments.
Once this is done, we should be able to add a line to our `.travis.yml` that enforces this for all future PRs.

mrd0ll4r:
We are now `golint`, `go vet` and `gosimple` clean.
I also fixed the starting of the prometheus server, simplified the http start/stop methods and implemented some things in the store start/stop methods.
The infohash middleware tests are nicer too.

This fixes #46 for the rewrite.

### DIFF
--- a/server/prometheus/prometheus.go
+++ b/server/prometheus/prometheus.go
@@ -71,6 +71,9 @@ type Server struct {
 
 var _ server.Server = &Server{}
 
+// Start starts the prometheus server and blocks until it exits.
+//
+// It panics if the server exits unexpectedly.
 func (s *Server) Start() {
 	s.grace = &graceful.Server{
 		Server: &http.Server{
@@ -93,6 +96,7 @@ func (s *Server) Start() {
 	log.Println("Prometheus server shut down cleanly")
 }
 
+// Stop stops the prometheus server and blocks until it exits.
 func (s *Server) Stop() {
 	s.grace.Stop(s.cfg.ShutdownTimeout)
 	<-s.grace.StopChan()
--- a/server/prometheus/prometheus.go
+++ b/server/prometheus/prometheus.go
@@ -8,6 +8,8 @@ package prometheus
 
 import (
 	"errors"
+	"log"
+	"net"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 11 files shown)


server/store/middleware/infohash/blacklist_test.go
```diff
@@ -10,41 +10,42 @@ import (
...
-				Name: "memory",
-			},
-			IPStore: store.DriverConfig{
-				Name: "memory",
-			},
-			PeerStore: store.DriverConfig{
-				Name: "memory",
-			},
-		},
+	mustGetStore = func() store.StringStore {
```
mrd0ll4r: I could pull this into the test functions and get rid of the `TestASetup` function, should I?

server/server.go
```diff
@@ -46,6 +46,11 @@ func New(cfg *chihaya.ServerConfig, tkr *tracker.Tracker) (Server, error) {
 
 // Server represents one instance of a server accessing the tracker.
 type Server interface {
+	// Start starts a server and blocks until the server exits.
+	//
+	// It panics if the server exits unexpectedly.
```
jzelinskie: It _should_ panic

server/prometheus/prometheus.go
```diff
@@ -80,8 +85,19 @@ func (s *Server) Start() {
...
 	}
+
+	if err := s.grace.ListenAndServe(); err != nil {
+		if opErr, ok := err.(*net.OpError); !ok || (ok && opErr.Op != "accept") {
+			log.Printf("Failed to gracefully run Prometheus server: %s", err.Error())
+			panic(err)
+		}
+	}
+
+	log.Println("Prometheus server shut down cleanly")
```
jzelinskie: Can this hunk be added in its own commit?

server/store/middleware/infohash/blacklist.go
```diff
@@ -13,17 +13,22 @@ import (
...
+	mustGetStore = func() store.StringStore {
+		return store.MustGetStore()
+	}
 }
 
 // ErrBlockedInfohash is returned by a middleware if any of the infohashes
 // contained in an announce or scrape are disallowed.
 var ErrBlockedInfohash = tracker.ClientError("disallowed infohash")
 
+var mustGetStore func() store.StringStore
```
jzelinskie: What are you getting out of making this a function and setting it as a global variable?

mrd0ll4r: I can mock the string store and don't need the whole store server running to test.

jzelinskie: Interesting. I think it might be _more_ boilerplate for middleware to implement this interface than to just initialize the whole server.

mrd0ll4r: I thought about putting the mock somewhere else, I just don't know where and if it's really useful

PR #163, opened 2016-04-07
Update README.md

gernest:
Fix installation steps.  The right path to the project root is `chihaya/src/github.com/chihaya/chihaya` on which `godep restore` is supposed to be executed.


jzelinskie:
Thanks for spotting this issue. I've updated our release notes, but I'm not going to merge this PR because all of our development is on the `middleware-refactor` branch which is based on master. I think we'll be merging it into master soon.

PR #165, opened 2016-04-08
middleware: added peer store updates middleware (swarm)

mrd0ll4r:


server/store/middleware/response/response.go
```diff
@@ -37,6 +40,48 @@ func responseAnnounceClient(next tracker.AnnounceHandler) tracker.AnnounceHandle
 			return err.(FailedToRetrievePeers)
```
jzelinskie: This should be `FailedToRetrievePeers(err)`. Woops, how did that get by?

server/store/middleware/response/response.go
```diff
@@ -37,6 +40,48 @@ func responseAnnounceClient(next tracker.AnnounceHandler) tracker.AnnounceHandle
...
 		}
 
+		switch req.Event {
```
jzelinskie: I think this should be before we get our peers (to reduce the amount of time that the resp.IPv4Peers and resp.IPv6Peers are in memory.

jzelinskie:
I played around with implementing this a bit this weekend.

I think there are some weird corner cases not present in here. I need to take a closer look.

Also, I'd pull most of it out into a function and then in the main middleware do something like

```go

if peer.IPv4 != nil {
  updatePeer(req, req.Peer())
}

if peer.IPv6 != nil {
  updatePeer(req, req.Peer6())
}

```

where `req.Peer` and `req.Peer6` are helpers to get the peer and peer6 variables

server/store/middleware/swarm/swarm.go
```diff
@@ -0,0 +1,65 @@
...
+func announceSwarmInteraction(next tracker.AnnounceHandler) tracker.AnnounceHandler {
+	return func(cfg *chihaya.TrackerConfig, req *chihaya.AnnounceRequest, resp *chihaya.AnnounceResponse) (err error) {
+		if req.IPv4 != nil {
+			err = updatePeerStore(req, req.Peer4())
+		}
+		if req.IPv6 != nil {
+			err = updatePeerStore(req, req.Peer6())
+		}
+
+		if err != nil {
```
jzelinskie: If a peer has both IPv4 and IPv6 and fails on IPv4, but succeeds on IPv6, this error is going to be nil and will be ignoring the first error.

server/store/middleware/swarm/swarm.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+	default:
+		if req.Left > 0 {
+			err = storage.PutLeecher(req.InfoHash, peer)
+		} else {
+			err = storage.PutSeeder(req.InfoHash, peer)
+		}
+	}
+
+	return
```
jzelinskie: here's the code I had mocked up to do similar to this:

```go
func handleEvent(req *chihaya.AnnounceRequest, peer chihaya.Peer) (err error) {
	storage := store.MustGetStore()

	switch {
	case ann.Event == event.Stopped || ann.Event == event.Paused:
		err = storage.DeleteSeeder(req.Infohash, peer)
		if err != nil {
			return err
		}

		err = storage.DeleteLeecher(req.Infohash, peer)
		if err != nil {
			return err
		}

	case ann.Event == event.Completed || req.Left == 0:
		err = storage.GraduateLeecher(req.Infohash, peer)
		if err != nil {
			return err
		}
	}
}
```

In this function I assumed that Delete wouldn't raise an error if you attempted to delete something that was not present. So to be safe, I actually remove the Peer from both seeders and leechers no matter what (since we could've missed the completed event and had them stored as a leecher previously when they're now a seeder). Continuing this same logic, I make completed and all requests with left == 0 use GraduateLeecher since it will guarantee that the peer is no longer in the leechers list if they're a seeder. I haven't looked at this code in a bit, but it looks like if we add a default case that simply sets the peer as a leecher, this should handle all cases.

mrd0ll4r: Yeah that looks pretty good, I would leave out the paused event though - we don't support the partial seeds extension to scrapes and I think we'd have to handle even more corner cases if we handle it.

mrd0ll4r:
This now depends on #170 for the ErrDNE

server/store/peer_store.go
```diff
@@ -27,7 +27,10 @@ type PeerStore interface {
 
 	// GraduateLeecher promotes a peer from a leecher to a seeder for the
 	// infoHash within the PeerStore.
+	// If the given Peer is not a leecher, it will still be added to the
```
jzelinskie: Can we add a blank comment line above this one?

jzelinskie:
You should add ErrDNE in this PR and remove it from #170 and then I'll merge this.

mrd0ll4r:
Moved ErrDNE over here, github shows a different ordering than what I see locally, not sure what's going on.

jzelinskie:
LGTM! Thanks for doing the cherrypick!

PR #166, opened 2016-04-12
Add peer parameters to AnnouncePeers

mrd0ll4r:


server/store/memory/peer_store.go
```diff
@@ -271,8 +271,14 @@ func (s *peerStore) AnnouncePeers(infoHash chihaya.InfoHash, seeder bool, numWan
 				}
 
 				if p.IP.To4() == nil {
+					if p.Equal(&peer6) {
+						continue
```
jzelinskie: numWant is never getting decremented when you do this continue.

mrd0ll4r: That's the idea - I don't add the announcing peer

jzelinskie: Oops -- was reading wayyy too fast. Should stop doing chihaya review during the workday 🙃

jzelinskie:
LGTM

PR #167, opened 2016-04-16
make peer Equal work on values

mrd0ll4r:
A non-pointer receiver and parameter make more sense to me.

Also formatted some stuff.


jzelinskie:
LGTM

PR #168, opened 2016-04-18
config: make example config work

mrd0ll4r:
The example config can now be used without making a change to get a public HTTP tracker.

This assumes #165 to be merged.

### DIFF
--- a/config_example.yaml
+++ b/config_example.yaml
@@ -7,16 +7,21 @@ chihaya:
     announce: 10m
     min_announce: 5m
     announce_middleware:
-      # These are currently fake values
-      - name: prometheus
-      - name: store_client_validation
-        config:
-          be_clever: true
-      - name: store_create_on_announce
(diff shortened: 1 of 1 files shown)


config_example.yaml
```diff
@@ -7,16 +7,21 @@ chihaya:
...
-      - name: store_create_on_announce
+#      - name: ip_blacklist
+#      - name: ip_whitelist
+#      - name: client_blacklist
+#      - name: client_whitelist
+#      - name: infohash_blacklist
+#      - name: infohash_whitelist
+#      - name: varinterval
+#      - name: deniability
+      - name: store_swarm_interaction
```
jzelinskie: This doesn't exist yet.

I also poked around with implementing this a bit this weekend and found we need to add some methods to our peer store.

jzelinskie:
LGTM

PR #169, opened 2016-04-21
make infohash and peerID byte arrays

mrd0ll4r:


pushrax:
It's guaranteed to be a 20-byte string right?

joshdekock:
@pushrax it says 20-byte in the spec, so I assume that all complying clients would use 20-bytes.

mrd0ll4r:
fixed, now says 20-byte string. (We don't actually check/enforce that yet, but we will soon)

jzelinskie:
If it is actually 20 bytes, why can't this type `[20]byte`?

mrd0ll4r:
I kinda want to make it a `[20]byte`, dunno how often we need to convert it to a string - I'll have a look.

server/store/middleware/infohash/blacklist_test.go
```diff
@@ -40,12 +40,26 @@ var mock store.StringStore = &storeMock{
 	strings: make(map[string]struct{}),
 }
 
+var (
+	ih1, ih2 chihaya.InfoHash
+)
+
+func init() {
```
jzelinskie: This doesn't need to be in an `init` function. You can use literals on new types derived from arrays: https://play.golang.org/p/s1ISSQ_XMq

jzelinskie:
I really dislike having this all over the place:

```go
id := [20]byte{}
copy(id[:], p.peerID)
```
Can we have a helper function that handles this?

chihaya.go
```diff
@@ -12,10 +12,50 @@ import (
...
+	copy(a[:], b)
+	return PeerID(a)
+}
+
+func PeerIDFromString(s string) PeerID {
+	if len(s) != 20{
+		panic("peer ID must be 20 bytes")
+	}
+
+	a := [20]byte{}
```
jzelinskie: Let's replace this line with `var buf [20]byte`.
For initializing a zero value, var is usually preferred.

chihaya.go
```diff
@@ -12,10 +12,50 @@ import (
...
+	if len(b) != 20 {
+		panic("peer ID must be 20 bytes")
+	}
+
+	a := [20]byte{}
+	copy(a[:], b)
+	return PeerID(a)
+}
+
+func PeerIDFromString(s string) PeerID {
```
jzelinskie: All these new functions should have doc strings.

jzelinskie:
LGTM

PR #170, opened 2016-04-22
add API server

mrd0ll4r:
This is my take on the API server. There are a few open points:

- should the API handle prefixes or should the user handle prefixes?
- do client IDs need to be urlEscaped?
- do we allow a middleware to register handlers for the API?
- should stores return errors for
    - does not exist for deletes?
    - does already exist for inserts?
- should adding/removing an infohash echo back a "sane" version of the infohash, e.g. in hex?

This is WIP, needs discussion.

### DIFF
--- a/server/store/middleware/client/routes.go
+++ b/server/store/middleware/client/routes.go
@@ -1,0 +1,85 @@
+package client
+
+import (
+	"errors"
+	"net/http"
+	"sync"
+
+	"github.com/julienschmidt/httprouter"
+
+	"net/url"
+
+	"github.com/chihaya/chihaya/server/store"
+)
+
+// PrefixClient is the prefix to be used for client peer IDs.
+const PrefixClient = "c-"
+
+const pathClient = "/clients/:client"
+
+var routesActivated sync.Once
+
+func activateRoutes() {
+	store.ActivateRoute(http.MethodPut, pathClient)
+	store.ActivateRoute(http.MethodDelete, pathClient)
+	store.ActivateRoute(http.MethodGet, pathClient)
+}
+
+func handleGetClient(w http.ResponseWriter, r *http.Request, p httprouter.Params) (int, interface{}, error) {
+	client, err := getClient(p)
+	if err != nil {
+		return http.StatusBadRequest, nil, err
+	}
+
+	match, err := store.MustGetStore().HasString(PrefixClient + client)
+	if err != nil {
+		panic(err)
+	}
+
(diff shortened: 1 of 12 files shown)


jzelinskie:
- I don't think prefixes should be in the store and that the store API should do anymore than present the store interfaces via an HTTP API.
- We should probably add a error in the store for when a resource doesn't exist and update the memory driver to use it.
- Adding handlers to API seems like an interesting idea

server/store/memory/ip_store.go
```diff
@@ -140,8 +140,13 @@ func (s *ipStore) HasAllIPs(ips []net.IP) (bool, error) {
 func (s *ipStore) RemoveIP(ip net.IP) error {
 	s.Lock()
 	defer s.Unlock()
+	key := key(ip)
```
jzelinskie: let's calculate the key before we lock
woo micro-optimizations

server/store/middleware/client/routes.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/chihaya/chihaya/server/store"
+)
+
+// PrefixClient is the prefix to be used for client peer IDs.
+const PrefixClient = "c-"
+
+const pathPutClient = "/client/:client"
+const pathDeleteClient = "/client/:client"
+
+var routesActivated = &sync.Once{}
```
jzelinskie: Is there a reason why this isn't just `var routesActivated sync.Once`?

mrd0ll4r: Just me programming way too much C(++) lately, will fix...

mrd0ll4r:
Should the store API provide hasIP/hasNetwork, getSeeders/getLeechers, numSeeders/numLeechers,... as well?

jzelinskie:
I think it should.

This PR is looking pretty good. It's got a lot of good stuff going on, it just appears to be 85% complete.
If you used `TODO`s into your code, I'd be more willing to merge things that are still in progress.

Some tiny nitpicks: 
- REST resources should be plural i.e. /network**s**/:network, ip**s**/:ip.
- code blocks should use triple upticks instead of indentation
- Let's rename ErrDNE to `ErrResourceDoesNotExist`

mrd0ll4r:
I added a few of those, namely matching IPs, infohashes and clientIDs. I'm not sure about the endpoints for all the peer store methods, so I'll leave that out for now.

jzelinskie:
You need to make a new PR with this branch pointed at master

PR #171, opened 2016-04-24
http: now validates length of info_hash and peer_id

mrd0ll4r:


mrd0ll4r:
I'll solve this in #169

PR #173, opened 2016-05-20
travis: add golint, goimports and go vet

mrd0ll4r:
This adds goimports, go vet and golint to travis. It excludes the `/vendor/` directory, just in case we use that at some point.

Good build example:
https://travis-ci.org/mrd0ll4r/chihaya/jobs/131722168

Bad build examples:
https://travis-ci.org/mrd0ll4r/chihaya/jobs/131721602
https://travis-ci.org/mrd0ll4r/chihaya/jobs/131720725

Example with an unclean file in the `/vendor/` directory:
https://travis-ci.org/mrd0ll4r/chihaya/jobs/131720094

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,13 @@ go:
 sudo: false
 install:
 - go get -t ./...
+- go get -u github.com/golang/lint/golint
+- go get -u golang.org/x/tools/cmd/goimports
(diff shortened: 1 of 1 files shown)


jzelinskie:
LGTM.

Thanks again!

PR #176, opened 2016-05-25
store: add general benchmark and test, add synchronization to string store

mrd0ll4r:
This is a WIP for improving  store synchronization, benchmarking and tests.

What was changed:
- The string store is now correctly started/stopped:
    - The constructor now accepts a shutdown channel and a wait group in addition to the config
    - When the channel is closed, the driver should shut down and notify the wait group
- The store package now has a general string store test that tests according to the interface
    - The memory implementation uses that test now
- The store package now has general string store benchmarks, that
    - create a new instance of the driver (clean start)
    - run a benchmark on it
    - close the driver, wait for it to close
- There are two kinds of benchmarks:
    - "normal" benchmarks that use the same value over and over again
    - "1K" benchmarks that use 1000 different values
    - This is to show caching effects
- The memory implementation now uses these benchmarks
- Uses of `testify.assert.*` were replaced with `testify.require.*` so that they fail the tests instead of just logging, continuing and failing at the end

Pro:
- Drivers can be implemented and tested much easier
- We have a way to compare performance of driver implementations

Con:
- The tests and benchmarks can not be in `_test.go` files due to visibility

Needs discussion:
- Too many benchmarks?
- Store stopping is pretty dumb - it just closes the channel and hopes for the best. Maybe we need an explicit `Stop` method for each driver that returns an error or something?
- Use `assert` over `require`?
- Make the amount of values in the `1K` benchmarks configurable?

This will fail travis because it has no documentation. Also needs some more formatting and whatnot.

### DIFF
--- a/pkg/stopper/stopper.go
+++ b/pkg/stopper/stopper.go
@@ -1,0 +1,101 @@
+// Copyright 2016 The Chihaya Authors. All rights reserved.
+// Use of this source code is governed by the BSD 2-Clause license,
+// which can be found in the LICENSE file.
+
+package stopper
+
+import (
+	"sync"
+)
+
+// AlreadyStopped is a closed error channel to be used by StopperFuncs when
+// an element was already stopped.
+var AlreadyStopped <-chan error
+
+// AlreadyStoppedFunc is a StopperFunc that returns AlreadyStopped.
+var AlreadyStoppedFunc = func() <-chan error { return AlreadyStopped }
+
+func init() {
+	closeMe := make(chan error)
+	close(closeMe)
+	AlreadyStopped = closeMe
+}
+
+// Stopper is an interface that allows a clean shutdown.
+type Stopper interface {
+	// Stop returns a channel that indicates whether the stop was
+	// successful.
+	// The channel can either return one error or be closed. Closing the
+	// channel signals a clean shutdown.
+	// The Stop function should return immediately and perform the actual
+	// shutdown in a seperate goroutine.
+	Stop() <-chan error
+}
+
+// StopGroup is a group that can be stopped.
+type StopGroup struct {
+	stoppables     []StopperFunc
+	stoppablesLock sync.Mutex
+}
+
+// StopperFunc is a function that can be used to provide a clean shutdown.
+type StopperFunc func() <-chan error
+
+// NewStopGroup creates a new StopGroup.
+func NewStopGroup() *StopGroup {
+	return &StopGroup{
+		stoppables: make([]StopperFunc, 0),
+	}
+}
+
+// Add adds a Stopper to the StopGroup.
+// On the next call to Stop(), the Stopper will be stopped.
+func (cg *StopGroup) Add(toAdd Stopper) {
+	cg.stoppablesLock.Lock()
+	defer cg.stoppablesLock.Unlock()
+
+	cg.stoppables = append(cg.stoppables, toAdd.Stop)
+}
+
+// AddFunc adds a StopperFunc to the StopGroup.
+// On the next call to Stop(), the StopperFunc will be called.
+func (cg *StopGroup) AddFunc(toAddFunc StopperFunc) {
+	cg.stoppablesLock.Lock()
+	defer cg.stoppablesLock.Unlock()
+
+	cg.stoppables = append(cg.stoppables, toAddFunc)
+}
+
+// Stop stops all members of the StopGroup.
+// Stopping will be done in a concurrent fashion.
+// The slice of errors returned contains all errors returned by stopping the
+// members.
+func (cg *StopGroup) Stop() []error {
+	cg.stoppablesLock.Lock()
+	defer cg.stoppablesLock.Unlock()
+
+	var errors []error
+	whenDone := make(chan struct{})
+
+	waitChannels := make([]<-chan error, 0, len(cg.stoppables))
+	for _, toStop := range cg.stoppables {
+		waitFor := toStop()
+		if waitFor == nil {
+			panic("received a nil chan from Stop")
+		}
+		waitChannels = append(waitChannels, waitFor)
+	}
+
+	go func() {
+		for _, waitForMe := range waitChannels {
+			err := <-waitForMe
+			if err != nil {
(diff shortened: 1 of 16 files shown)


jzelinskie:
* Your commit message is too long (see the commits page to see how it overflows). You can always add a paragraph below the first line.
* Benchmarks look good -- we should add a testing section to the README and include a brief bit about this.
* I like [this package](https://github.com/coreos/jwtproxy/blob/master/stop/stopper.go) for stopping stuff.  It calls the `Stop` method or a provided function asynchronously and waits for them all to return.
* :+1: to require over assert. I didn't know it existed.

server/store/ip_store.go
```diff
@@ -75,10 +75,10 @@ func RegisterIPStoreDriver(name string, driver IPStoreDriver) {
 }
 
 // OpenIPStore returns an IPStore specified by a configuration.
-func OpenIPStore(cfg *DriverConfig) (IPStore, error) {
+func OpenIPStore(cfg *DriverConfig) (IPStore, StopperFunc, error) {
```
jzelinskie: just make IPStore implement `Stop()`

server/store/stopper.go
```diff
@@ -0,0 +1,97 @@
+package store
```
jzelinskie: let's put stopper in a stopper package in the `pkg` directory

server/store/ip_store.go
mrd0ll4r: I'm not feeling good about this - store just embeds IPStore, StringStore and PeerStore, all of them would have a `Stop()` func, plus the store itself has one. I'm not familiar with how go handles that

jzelinskie: Does this answer your question? https://play.golang.org/p/sws9_zkqER

mrd0ll4r: It does! But now we expose the (IP|String|Peer)Store.Stop() func to every user of the store :(

jzelinskie: I'm not sure what the issue being discussed is anymore. Embedding an exported type being exposed is completely unrelated to how you can implement a `Stop` method on a type that calls the stop methods on it's contained members: https://play.golang.org/p/bd9bHfW8Gf

server/store/memory/ip_store_test.go
```diff
@@ -10,13 +10,15 @@ import (
...
 )
 
 var (
-	v6  = net.ParseIP("0c22:384e:0:0c22:384e::68")
-	v4  = net.ParseIP("12.13.14.15")
-	v4s = net.ParseIP("12.13.14.15").To4()
+	v6                 = net.ParseIP("0c22:384e:0:0c22:384e::68")
+	v4                 = net.ParseIP("12.13.14.15")
+	v4s                = net.ParseIP("12.13.14.15").To4()
+	ipStoreTester      = store.PrepareIPStoreTester(&ipStoreDriver{})
```
jzelinskie: Newline above this line to separate the indentation.

mrd0ll4r: fixed

pkg/stopper/stopper.go
```diff
@@ -0,0 +1,97 @@
+package stopper
```
jzelinskie: copyright notice at the top of files

server/store/README.md
```diff
@@ -0,0 +1,43 @@
...
+During runtime, each of them will be implemented by a driver.
+Even though all different drivers for one interface provide the same functionality, their behaviour can be very different.
+For example: The memory implementation keeps all state in-memory - this is very fast, but not persistent, it loses its state on every restart.
+A database-backed driver on the other hand could provide persistence, at the cost of performance.
+
+The pluggable design of Chihaya allows for the different interfaces to use different drivers.
+For example: A typical use case of the string store is to provide blacklists or whitelists for infohashes/client IDs/....
+You'd typically want these lists to be persistent, so you'd choose a driver that provides persistence.
+The peer store on the other hand rarely needs to be persistent, as all peer state will be restored after one announce interval.
+You'd therefore typically choose a very performant but non-persistent driver for the peer store.
```
jzelinskie: PeerStore

server/store/memory/peer_store_test.go
```diff
@@ -5,155 +5,144 @@
...
-		if p.seeder {
-			err = s.PutSeeder(hash, peer)
-		} else {
-			err = s.PutLeecher(hash, peer)
-		}
-		assert.Nil(t, err)
-	}
+var peerStoreTester = store.PreparePeerStoreTester(&peerStoreDriver{})
+var peerStoreBenchmarker = store.PreparePeerStoreBenchmarker(&peerStoreDriver{})
+var peerStoreBenchConf = &store.DriverConfig{}
```
jzelinskie:
```go
var (
  peerStoreTester = store.PreparePeerStoreTester(&peerStoreDriver{})
  peerStoreBenchmarker = store.PreparePeerStoreBenchmarker(&peerStoreDriver{})
  peerStoreBenchConf = &store.DriverConfig{}
)```

File: server/store/memory/string_store.go [hunk 7]
```diff
@@ -59,3 +60,7 @@ func (ss *stringStore) RemoveString(s string) error {
 
 	return nil
 }
+
+func (ss *stringStore) Stop() <-chan error {
+	return stopper.AlreadyStopped
```
jzelinskie: Are we sure we don't want this to throwaway the current state of the internal map of strings? Thus the next GC would free all of that memory. By not doing so, we're making an assumption about the lifetime of the program.

server/store/store_tests.go
```diff
@@ -0,0 +1,511 @@
+package store
```
jzelinskie: copyright notice

server/store/store_bench.go
```diff
@@ -0,0 +1,1257 @@
+package store
```
jzelinskie: copyright notice.... you get the picture

mrd0ll4r: huh yeah, I should make a checklist for this stuff

jzelinskie:
LGTM

PR #181, opened 2016-06-23
memory: add peer store gc

### LINKED ISSUE #179: Peer store doesn't actually collect garbage
In the example config we have a `gcAfter: 30m` for the peer store config. We should either remove that or make the peer store collect garbage.

mrd0ll4r:
This is to fix #179


mrd0ll4r:
#196 does this for the refactor branch, closing.

PR #187, opened 2016-07-13
config: add middleware configuration to example config

mrd0ll4r:


jzelinskie:
I'm going to close this, but we should keep the branch for reference until these middleware are implemented again.

PR #189, opened 2016-07-14
http: fix encoding of IPv4 peers with 16-byte addresses

mrd0ll4r:
This is a debatable fix/workaround if a PeerStore always returns a 16-byte address, even for IPv4 peers.

Pro: Now behaves the same way for compact and non-compact responses (we us `net.IP.String()` for non-compact, which also detects 16-byte IPv4 addresses)

Con: Not fixing the problem? Maybe we should add a test for PeerStores that verifies that IPv4 peers actually have 4-byte addresses?


mrd0ll4r:
#198 does this for the refactor branch, closing.

PR #190, opened 2016-07-20
store: do not test AnnouncePeers

mrd0ll4r:
I believe that we should not test whatever `AnnouncePeers` returns at all.

### DIFF
--- a/server/store/store_tests.go
+++ b/server/store/store_tests.go
@@ -510,10 +510,10 @@ func (pt *peerStoreTester) TestPeerStore(t *testing.T, cfg *DriverConfig) {
 	require.Equal(t, 7, s.NumSeeders(hash))
 	require.Equal(t, 3, s.NumLeechers(hash))
 
-	peers1, peers61, err := s.AnnouncePeers(hash, true, 5, peer, chihaya.Peer{})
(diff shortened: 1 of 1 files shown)


server/store/store_tests.go
```diff
@@ -510,10 +510,10 @@ func (pt *peerStoreTester) TestPeerStore(t *testing.T, cfg *DriverConfig) {
 	require.Equal(t, 7, s.NumSeeders(hash))
 	require.Equal(t, 3, s.NumLeechers(hash))
 
-	peers1, peers61, err := s.AnnouncePeers(hash, true, 5, peer, chihaya.Peer{})
+	_, _, err = s.AnnouncePeers(hash, true, 5, peer, chihaya.Peer{})
 	require.Nil(t, err)
-	require.NotNil(t, peers1)
-	require.NotNil(t, peers61)
+	// We can not test the returned slices. They change depending on the
```
jzelinskie: Rephrase this and put it above the `require.Nil`. Comments in Go always go above what they're describing.

mrd0ll4r: Alright, rephrased as good as possible :)

jzelinskie:
LGTM

PR #196, opened 2016-08-17
memory: fix garbage collection, add config check

mrd0ll4r:
GC now uses the `PeerLifetime` field.
I added the `<=0` check because that would send the goroutine to spin on the select. I don't really care if people set `PeerLifetime` to zero, may they collect all their peers...


jzelinskie:
LGTM!

PR #197, opened 2016-08-17
Refactor udp

mrd0ll4r:
A few fixes:

- the `bytepool` fix: newly created slices had a length of 256, recycled ones had `cap(b)==2048`
- the `t.wg.Wait()` fix: we already wait on `wg` in `Stop()`
- the `handleRequest` signature change: response slice is no longer used
- the `writer` fixes: make it so that the whole response is sent back instead of only the header :)
- some cleanup in the parser

### DIFF
--- a/frontend/udp/parser.go
+++ b/frontend/udp/parser.go
@@ -12,7 +12,6 @@ const (
 	announceActionID
 	scrapeActionID
 	errorActionID
-	announceDualStackActionID
 )
 
 // Option-Types as described in BEP 41 and BEP 45.
--- a/frontend/udp/frontend.go
+++ b/frontend/udp/frontend.go
@@ -6,7 +6,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
-	"log"
 	"net"
 	"sync"
 	"time"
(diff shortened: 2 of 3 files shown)


frontend/udp/parser.go
```diff
@@ -67,6 +66,7 @@ func ParseAnnounce(r Request, allowIPSpoofing bool) (*bittorrent.AnnounceRequest
 	ip := r.IP
 	ipbytes := r.Packet[84:88]
 	if allowIPSpoofing {
+		// TODO(mrd0ll4r) check if this is an empty IP address?
```
jzelinskie: To respond to your question earlier on IRC. I used logic based on the behavior of the HTTP frontend when removing this before. If you  have enable spoofing and an HTTP announce has `&ip=0.0.0.0`, we don't discount that.

mrd0ll4r: Hmm, I feel like the client doesn't know whether we have IP spoofing enabled or not and just sends 0.0.0.0 to tell us "use my public IP", but no idea if that's true or documented anywhere

jzelinskie:
LGTM

PR #198, opened 2016-08-18
http: fix encoding of IPv4 peers with 16-byte addresses

mrd0ll4r:
This is #189 for the refactor branch.

### DIFF
--- a/frontend/http/writer.go
+++ b/frontend/http/writer.go
@@ -36,7 +36,7 @@ func WriteAnnounceResponse(w http.ResponseWriter, resp *bittorrent.AnnounceRespo
 
 		// Add the IPv4 peers to the dictionary.
 		for _, peer := range resp.IPv4Peers {
-			IPv4CompactDict = append(IPv4CompactDict, compact(peer)...)
+			IPv4CompactDict = append(IPv4CompactDict, compact4(peer)...)
 		}
 		if len(IPv4CompactDict) > 0 {
 			bdict["peers"] = IPv4CompactDict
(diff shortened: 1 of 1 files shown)


jzelinskie:
I'd rather have an extra test for storage implementations than running the risk of mistakenly reformatted IPv4-compatible IPv6 addresses into IPv4 ones.

mrd0ll4r:
What about this: Let's introduce a `compact4` and a `compact6` function, to be called for `peers4` and `peers6` respectively. The former just calls `To4` on every IP, the latter calls `To16` ?

jzelinskie:
That sounds fair.

mrd0ll4r:
Alright, now using two functions, can you take another look please?

frontend/http/writer.go
```diff
@@ -82,8 +82,15 @@ func WriteScrapeResponse(w http.ResponseWriter, resp *bittorrent.ScrapeResponse)
 	})
 }
 
-func compact(peer bittorrent.Peer) (buf []byte) {
-	buf = []byte(peer.IP)
+func compact4(peer bittorrent.Peer) (buf []byte) {
+	buf = []byte(peer.IP.To4())
```
jzelinskie: In the off chance that peer.IP.To4() returns nil, this isn't going to fail, but will provide an invalid response.
Let's panic if it's nil.

mrd0ll4r: yeah I thought about that too, but was hesitant to panic. Still, added it now.

jzelinskie:
LGTM

PR #200, opened 2016-08-20
cmd/chihaya: clean up

mrd0ll4r:
Some cleaning up and fixes here and there.

I made a separate channel for signals, otherwise we could close it and send a signal on it while we are shutting down which would then panic.


jzelinskie:
LGTM. This file is getting a little overloaded and think once we start getting middleware initialization in it, we're going to have to break it into multiple files.

PR #201, opened 2016-08-24
cmd: pull closure out into its own func and add config.go

jzelinskie:
...wow the diff sucks for this. I basically just pulled the closure out into its own function that returns an error and then moved the config stuff into its own file.

### DIFF
--- a/cmd/chihaya/main.go
+++ b/cmd/chihaya/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"errors"
-	"io/ioutil"
 	"log"
 	"net/http"
 	"os"
@@ -12,7 +11,6 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/cobra"
-	"gopkg.in/yaml.v2"
 
 	httpfrontend "github.com/chihaya/chihaya/frontend/http"
 	udpfrontend "github.com/chihaya/chihaya/frontend/udp"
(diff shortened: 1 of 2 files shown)


cmd/chihaya/main.go
```diff
@@ -12,178 +11,138 @@ import (
...
+		pprof.StartCPUProfile(f)
+		defer pprof.StopCPUProfile()
 	}
 
-	f, err := os.Open(os.ExpandEnv(path))
+	configFilePath, _ := cmd.Flags().GetString("config")
+	configFile, err := ParseConfigFile(configFilePath)
 	if err != nil {
-		return nil, err
+		return errors.New("failed to read config: " + err.Error())
```
mrd0ll4r: We could start using github.com/pkg/errors and make this to `errors.Wrap(err,"failed to read config")`. I use the package for pretty much every new project I start because I have to wrap so often.

cmd/chihaya/main.go
```diff
@@ -12,178 +11,138 @@ import (
...
+			if err := rootCmdRun(cmd, args); err != nil {
 				log.Fatal(err)
 			}
 		},
 	}
-
-	rootCmd.Flags().StringVar(&configFilePath, "config", "/etc/chihaya.yaml", "location of configuration file")
-	rootCmd.Flags().StringVarP(&cpuProfilePath, "cpuprofile", "", "", "location to save a CPU profile")
+	rootCmd.Flags().String("config", "/etc/chihaya.yaml", "location of configuration file")
+	rootCmd.Flags().StringP("cpuprofile", "", "", "location to save a CPU profile")
```
mrd0ll4r: I just noticed: You made this a string parameter with `""` as the shorthand - intended? Why?
Also, let's give `config` a shorthand, now that I know about the feature :)

cmd/chihaya/main.go
jzelinskie: I'd be ok with this. Maybe we should do it as a separate PR.

cmd/chihaya/main.go
jzelinskie: This was definitely a mistake. I'm not sure what good shorthand characters would be a good choice.

cmd/chihaya/main.go
mrd0ll4r: Yup, sounds good

mrd0ll4r:
LGTM

PR #203, opened 2016-08-31
udp: implement bep41

mrd0ll4r:
I was thinking if we should maybe put the path in the `Params` as well (under some predefined key) - opinions?


frontend/udp/parser.go
```diff
@@ -99,41 +103,69 @@ func ParseAnnounce(r Request, allowIPSpoofing bool) (*bittorrent.AnnounceRequest
...
 
-			// TODO(chihaya): Actually parse the URL Data as described in BEP 41
-			// into something that fulfills the bittorrent.Params interface.
+			n, err := buf.Write(packet[i+2 : i+2+length])
+			if err != nil {
+				return nil, err
+			}
+			if n != length {
+				return nil, fmt.Errorf("expected to write %d bytes, wrote %d", length, n)
+			}
```
jzelinskie: Why don't we call parseURLData around here rather than at the multiple different return sites?

frontend/udp/parser.go
```diff
@@ -99,41 +103,69 @@ func ParseAnnounce(r Request, allowIPSpoofing bool) (*bittorrent.AnnounceRequest
...
-			if optionStartIndex+1+length > len(packet)-1 {
-				return params, errMalformedPacket
+			length := int(packet[i+1])
+			if i+2+length > len(packet) {
+				return nil, errMalformedPacket
 			}
 
-			// TODO(chihaya): Actually parse the URL Data as described in BEP 41
-			// into something that fulfills the bittorrent.Params interface.
+			n, err := buf.Write(packet[i+2 : i+2+length])
```
jzelinskie: why use a buffer when you could do `url := string(packet[i+2 : i+2+length])`?

frontend/udp/parser.go
```diff
@@ -99,41 +103,69 @@ func ParseAnnounce(r Request, allowIPSpoofing bool) (*bittorrent.AnnounceRequest
...
+	return "", false
+}
+
+var emptyURLData = &optionalURLData{}
+
+func parseURLData(urlData string) (bittorrent.Params, error) {
+	if urlData == "" {
+		return emptyURLData, nil
+	}
+	u, err := url.ParseRequestURI(urlData)
```
jzelinskie: This is pretty slow, which is why we don't use it for HTTP. Maybe we should look at reusing that here.

frontend/udp/parser_test.go
```diff
@@ -0,0 +1,144 @@
...
+import (
+	"bytes"
+	"testing"
+)
+
+func TestParseQuery(t *testing.T) {
+	passkey := "abc123"
+	buf := &bytes.Buffer{}
+
+	//now add optional parameters
```
jzelinskie: All of these tests are very similar in structure. They could probably be turned into a [table-driven tests](https://github.com/golang/go/wiki/TableDrivenTests) where the table looks like
```go
struct {
  header []byte
  urlData string
}{
{ []byte{0x2, 0x10}, "/?passkey=abc123" },
{ []byte{0x2, 0x0}, "" },
...
}
```

frontend/udp/parser.go
mrd0ll4r: We can have multiple blocks of URLData (for path+querystrings longer than 255 bytes), so we have to collect them all first, then parse at the end.

frontend/udp/parser.go
mrd0ll4r: Yeah I thought about that too. I used the stdlib here for feature-completeness and robustness. Keep in mind that we can also have paths in there, i.e. `/some/path?with=a&query=string`.

frontend/udp/parser_test.go
mrd0ll4r: Yeah that's true. Will do

frontend/udp/parser.go
mrd0ll4r: To concatenate multiple URLData blocks. Then again I don't know how often you'd have multiple blocks of URLData anyways, so using them as strings and concatenating in the improbable case might be the better choice.

We could pool the buffers?

jzelinskie: Is there somewhere in the spec or real life where you've seen multiple URL data blocks?

mrd0ll4r: BEP41, in the section on Option Types, the bullet point for URLData: "If this option appears more than once, the data fields are concatenated. This allows clients to send PATH and QUERY strings that are longer than 255 bytes, chunked into blocks of no larger than 255 bytes."

jzelinskie: Cool. I think you should also run `-gcflags="-m"` to determine if the buffer escapes or if it'll remain on the stack.

mrd0ll4r: Yup:

```
./parser.go:130: buf escapes to heap
./parser.go:111: moved to heap: buf
```

So what's the verdict? Pool the buffer?

jzelinskie: yea, but now we need to pass a bufferpool around or make it a global >_<

mrd0ll4r: I made a global free list of buffers, the same way the `fmt` package does. We should be able to use the same pool for the buffers in `writer.go`, but I'll do that in another PR.

jzelinskie:
Maybe our query parser should be moved into the `bittorrent` package and become a constructor for Params and then add a `Raw()` method for those that want the path.

mrd0ll4r:
I agree we should pull it out of HTTP to reuse it for UDP, but the `bittorrent` package feels out of place, what has an URL parser to do with BitTorrent? Maybe make a `urlparse`/`queryparse` package parallel to the `stopper` package?

jzelinskie:
Moving it all into `params.go` inside the bittorrent package makes sense to me. It's not general purpose enough to stand on its own since it makes quite a few tradeoffs in terms of optimizing behavior specific to the properties of BitTorrent announces i.e. last value wins except for multiple infohashes for scrapes.

mrd0ll4r:
Query parsing is now in the `bittorrent` package with a few minor changes to make it work the same way for HTTP and UDP. Can you take another look please?

bittorrent/params.go
```diff
@@ -0,0 +1,193 @@
...
+// Note that, in the case of a key occurring multiple times in the query, only
+// the last value for that key is kept.
+// The only exception to this rule is the key "info_hash" which will attempt to
+// parse each value as an InfoHash and return an error if parsing fails. All
+// InfoHashes are collected and can later be retrieved by calling the InfoHashes
+// method.
+func ParseURLData(urlData string) (*QueryParams, error) {
+	var (
+		path, query string
+	)
```
jzelinskie: rewrite this block to be one line
```go
var path, query string
```

jzelinskie:
Besides that one tiny style nitpick. LGTM!

mrd0ll4r:
Both PRs updated - if you want to merge, merge this one first, I'll have to update #206

PR #207, opened 2016-09-03
merge refactor into master

jzelinskie:
There's not much left before this branch is more stable than master and it's definitely the head of development.

Why not just merge now?


.travis.yml
```diff
@@ -11,7 +10,7 @@ script:
 - go test -v $(go list ./... | grep -v /vendor/)
 - go vet $(go list ./... | grep -v /vendor/)
 - diff <(goimports -d $(find . -type f -name '*.go' -not -path "./vendor/*")) <(printf "")
-- for d in $(go list ./... | grep -v /vendor/); do diff <(golint $d) <(printf ""); done
+#- for d in $(go list ./... | grep -v /vendor/); do diff <(golint $d) <(printf ""); done
```
mrd0ll4r: I think we should quickly document the missing things and get lint clean before merging

mrd0ll4r:
We don't have the final hook(s) and/or whatever handles the storage

bittorrent/bittorrent.go
```diff
@@ -146,16 +119,20 @@ type Peer struct {
...
+type ClientError string
+
+// Error implements the error interface for ClientError.
+func (c ClientError) Error() string { return string(c) }
 
-// Params is used to fetch request parameters.
-type Params interface {
-	String(key string) (string, error)
+// Tracker represents an implementation of the BitTorrent tracker protocol.
+type Tracker interface {
```
mrd0ll4r: Do we actually use this anywhere?

example_config.yaml
```diff
@@ -0,0 +1,40 @@
...
+      jwk_set_update_interval: 5m
+  - name: client approval
+    config:
+      whitelist:
+      - OP1011
+      blacklist:
+      - OP1012
+
+  posthooks:
+  - name: gossip
```
mrd0ll4r: I think we should comment this out for the time being?

frontend/http/frontend.go
```diff
@@ -0,0 +1,179 @@
...
+	start := time.Now()
+	defer recordResponseDuration("announce", err, time.Since(start))
+
+	req, err := ParseAnnounce(r, t.RealIPHeader, t.AllowIPSpoofing)
+	if err != nil {
+		WriteError(w, err)
+		return
+	}
+
+	resp, err := t.logic.HandleAnnounce(context.TODO(), req)
```
mrd0ll4r: We're still using `context.TODO()` for these - we could simply use `context.Background` I think

frontend/http/frontend.go
```diff
@@ -0,0 +1,179 @@
...
+		return
+	}
+
+	err = WriteAnnounceResponse(w, resp)
+	if err != nil {
+		WriteError(w, err)
+		return
+	}
+
+	go t.logic.AfterAnnounce(context.TODO(), req, resp)
```
mrd0ll4r: Same as above

frontend/http/writer_test.go
```diff
@@ -1,15 +1,12 @@
...
-// which can be found in the LICENSE file.
-
 package http
 
 import (
 	"net/http/httptest"
 	"testing"
 
-	"github.com/chihaya/chihaya/tracker"
 	"github.com/stretchr/testify/assert"
```
mrd0ll4r: We should either use `..testify/require` to actually make the tests fail or use the stdlib testing everywhere (I know you didn't use `assert` or `require` in most of the new stuff, so I'm bringing this up for consistency)

frontend/udp/bytepool/bytepool.go
```diff
@@ -0,0 +1,34 @@
...
+
+import "sync"
+
+// BytePool is a cached pool of reusable byte slices.
+type BytePool struct {
+	sync.Pool
+}
+
+// New allocates a new BytePool with slices of the provided capacity.
+func New(length, capacity int) *BytePool {
```
mrd0ll4r: Tbh I think we should only specify length. If people use different values for `length` and `capacity` (let's say 1024 and 4096) the length of slices that are pooled will be different from slices that are newly allocated because we reslice them to `[:cap(b)]` when we return them to the pool.

frontend/udp/frontend.go
```diff
@@ -0,0 +1,246 @@
...
+	case connectActionID:
+		actionName = "connect"
+
+		if !bytes.Equal(connID, initialConnectionID) {
+			err = errMalformedPacket
+			return
+		}
+
+		WriteConnectionID(w, txID, NewConnectionID(r.IP, time.Now(), t.PrivateKey))
+		return
```
mrd0ll4r: We have all these unnecessary `return`s in the `switch` oO

bittorrent/bittorrent.go
jzelinskie: nope!

example_config.yaml
jzelinskie: It's fine for now, it doesn't break anything and I like having config first to drive the design when you go to write it.

frontend/http/frontend.go
jzelinskie: agreed

frontend/http/writer_test.go
jzelinskie: Yea, it's up to you. I don't really care too much, since all it does is eliminate a small amount of boilerplate.

frontend/udp/bytepool/bytepool.go
jzelinskie: Sounds good

jzelinskie:
I don't think we need to change the signatures to return ctx, because context is an interface (aka pointer).

frontend/http/writer_test.go
mrd0ll4r: Alright, let's not make changes now and later upgrade them to `require` all at once.

mrd0ll4r:
LGTM from my point of view, nice!

PR #208, opened 2016-09-03
udp: use free list for response buffers

mrd0ll4r:


frontend/udp/writer.go
```diff
@@ -47,35 +47,38 @@ func WriteAnnounce(w io.Writer, txID []byte, resp *bittorrent.AnnounceResponse,
...
-		binary.Write(&buf, binary.BigEndian, scrape.Complete)
-		binary.Write(&buf, binary.BigEndian, scrape.Snatches)
-		binary.Write(&buf, binary.BigEndian, scrape.Incomplete)
+		binary.Write(buf, binary.BigEndian, scrape.Complete)
+		binary.Write(buf, binary.BigEndian, scrape.Snatches)
+		binary.Write(buf, binary.BigEndian, scrape.Incomplete)
 	}
 
 	w.Write(buf.Bytes())
+	buf.free()
```
jzelinskie: shouldn't we we be defering this?

mrd0ll4r: I don't see why, should be good like this?

jzelinskie:
LGTM

PR #217, opened 2016-09-07
bencode: add missing error handling

jzelinskie:
Go report card runs `gocyclo -n 15`, while trying to reduce the complexity of some of our calls, I noticed that the bencode encoder is ignoring all the errors from calling `Write`. This is fine if you're working with bytes.Buffer or memory, but if you were to use another instance of an io.Writer, it could fail silently.

### DIFF
--- a/frontend/http/bencode/encoder.go
+++ b/frontend/http/bencode/encoder.go
@@ -25,8 +25,8 @@ func (enc *Encoder) Encode(v interface{}) error {
 
 // Marshal returns the bencoding of v.
 func Marshal(v interface{}) ([]byte, error) {
-	buf := &bytes.Buffer{}
-	err := marshal(buf, v)
+	var buf bytes.Buffer
+	err := marshal(&buf, v)
 	return buf.Bytes(), err
 }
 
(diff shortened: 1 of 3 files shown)


frontend/http/bencode/encoder.go
```diff
@@ -36,124 +36,154 @@ type Marshaler interface {
...
+	case []byte:
+		err = marshalBytes(w, v)
 
 	case string:
-		marshalString(w, v)
+		err = marshalString(w, v)
 
-	case int:
-		marshalInt(w, int64(v))
+	case []string:
```
mrd0ll4r: Why do we have an extra case for this and not let it be handled by `[]interface{}` ?

jzelinskie: IIRC some of these extra definitions are for performance, but I doubt it makes any difference. I'll squash it down.

jzelinskie: Apparently, you can't do this.

mrd0ll4r:
LGTM

PR #218, opened 2016-09-07
travis: fix swallowing of exit codes in golint loop

mrd0ll4r:
This should fix the non-failed-but-should-have-failed travis build (https://travis-ci.org/chihaya/chihaya/builds/158065317)

The change in bencode is to see if travis fails now, but it might not use the .travis.yml from the PR. Let's see.


jzelinskie:
LGTM

PR #219, opened 2016-09-08
readme: update diagram

jzelinskie:
I have to say: it's awesome being able to see a diff of the diagrams!


mrd0ll4r:
Yup, and a good choice on the diagram! LGTM :)

PR #220, opened 2016-09-08
Storage scrape support

mrd0ll4r:


storage/memory/peer_store.go
```diff
@@ -92,6 +92,14 @@ func (s *peerStore) shardIndex(infoHash bittorrent.InfoHash, p bittorrent.Peer)
 	return idx
 }
 
+func (s *peerStore) shardIndexByAddressType(infoHash bittorrent.InfoHash, v6 bool) uint32 {
```
jzelinskie: Couldn't we merge this with the function above and and put `len(p.IP) == net.IPv6len` in all the call sites?

Also, I'm not totally positive that I understand how incrementing the index this way avoids collisions.

mrd0ll4r: Yeah I thought about merging/chaining them, I'll do it.

About the index: We don't do that to avoid collisions - we simply use the first half of all shards for IPv4 swarms and the second half for IPv6 swarms.

jzelinskie: Shouldn't they both be `/2` then?

mrd0ll4r: Oh there is definitely one too many plus signs in there, that's true...
And yeah, the modulo should be with `len/2`, whoops.

storage/memory/peer_store.go
```diff
@@ -84,10 +84,12 @@ type peerStore struct {
 
 var _ storage.PeerStore = &peerStore{}
 
-func (s *peerStore) shardIndex(infoHash bittorrent.InfoHash, p bittorrent.Peer) uint32 {
-	idx := binary.BigEndian.Uint32(infoHash[:4]) % uint32(len(s.shards))
-	if len(p.IP) == net.IPv6len {
-		idx += idx + uint32(len(s.shards)/2)
+func (s *peerStore) shardIndex(infoHash bittorrent.InfoHash, v6 bool) uint32 {
+	// we actually have twice the amount of shards specified by the user:
```
jzelinskie: I've been trying my best not to ever write "we" or "our" in documentation.

"There are twice the amount of shards specified by the user, the first half is dedicated to IPv4 swarms and the second half is dedicated to IPv6 swarms"

jzelinskie:
LGTM

PR #226, opened 2016-09-24
cmd/chihaya: stop hooks

### LINKED ISSUE #214: need to safely shutdown hooks in main
https://github.com/chihaya/chihaya/blob/6b376e35221ac4c31a645e5e47617b33192e6314/cmd/chihaya/main.go#L122

mrd0ll4r:
Fixes #214.

### DIFF
--- a/cmd/chihaya/main.go
+++ b/cmd/chihaya/main.go
@@ -119,7 +119,11 @@ func rootCmdRun(cmd *cobra.Command, args []string) error {
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


jzelinskie:
I think the stopping logic could better be located as a Stop method on middleware.Logic.

mrd0ll4r:
Sure, that works. We'll have to create a new Logic when we re-parse the config, but that should be fine and we probably have to do that anyway.

jzelinskie:
LGTM!

PR #227, opened 2016-09-25
jwt: add updateKeys method and call in constructor

### LINKED ISSUE #225: JWT middleware needs to get initial JWT set
(Untested, from code only)

The JWT middleware works with an empty set of keys until it runs the update logic in the goroutine for the first time. We should probably just attempt to update in the constructor, before starting the goroutine.

jzelinskie:
Fixes #225.

### DIFF
--- a/cmd/chihaya/config.go
+++ b/cmd/chihaya/config.go
@@ -78,7 +78,11 @@ func (cfg ConfigFile) CreateHooks() (preHooks, postHooks []middleware.Hook, err 
 			if err != nil {
 				return nil, nil, errors.New("invalid JWT middleware config: " + err.Error())
 			}
-			preHooks = append(preHooks, jwt.NewHook(jwtCfg))
+			hook, err := jwt.NewHook(jwtCfg)
+			if err != nil {
+				return nil, nil, errors.New("invalid JWT middleware config: " + err.Error())
+			}
+			preHooks = append(preHooks, hook)
 		case "client approval":
 			var caCfg clientapproval.Config
 			err := yaml.Unmarshal(cfgBytes, &caCfg)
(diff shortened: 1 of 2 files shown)


middleware/jwt/jwt.go
```diff
@@ -56,44 +56,49 @@ func NewHook(cfg Config) middleware.Hook {
...
-				h.publicKeys = keys
+				h.updateKeys()
 			}
 		}
 	}()
 
 	return h
 }
 
+func (h *hook) updateKeys() {
```
mrd0ll4r: Make this return an `error` - I believe we want to know if it fails (at least for the initial update).

jzelinskie: fair point

cmd/chihaya/config.go
```diff
@@ -78,7 +78,11 @@ func (cfg ConfigFile) CreateHooks() (preHooks, postHooks []middleware.Hook, err
 			if err != nil {
 				return nil, nil, errors.New("invalid JWT middleware config: " + err.Error())
 			}
-			preHooks = append(preHooks, jwt.NewHook(jwtCfg))
+			hook, err := jwt.NewHook(jwtCfg)
+			if err != nil {
+				return nil, nil, errors.New("failed to start JWT middleware server: " + err.Error())
```
mrd0ll4r: I wouldn't call this a server, it's not serving anything.

jzelinskie: What should we call it then? It sits in the background updating the JWK set.

mrd0ll4r: I wouldn't assume anything about what failed during the attempt to construct the JWT middleware, I'd just say "failed to start JWT middleware" or something like that

jzelinskie:
ptal

PR #234, opened 2016-09-28
storage: react correctly to ErrDNE for announces

### LINKED ISSUE #231: Announce responses must always contain the announcing peer
We have to implement this somewhere - probably in the response middleware. (Because the store, as of now, returns `ErrResourceDoesNotExist` if it doesn't track the swarm.)

### LINKED ISSUE #232: `storage.AnnouncePeers` returns `ErrResourceDoesNotExist`, which then causes other problems
Vague title, the point is: The storage returns `ErrResourceDoesNotExist`, the logic sees the error and fails the announce, no gets added to the swarm, all future Announces also produce the error, ...

I think we should fix this in the `responseHook` (check for `ErrResourceDoesNotExist`) and document the `storage` interface accordingly (nothing is said about that error being returned).

mrd0ll4r:
Fixes #232
Fixes #231

### DIFF
--- a/middleware/hooks.go
+++ b/middleware/hooks.go
@@ -102,14 +102,30 @@ func (h *responseHook) HandleAnnounce(ctx context.Context, req *bittorrent.Annou
 
 	switch len(req.IP) {
 	case net.IPv4len:
-		resp.IPv4Peers, err = h.store.AnnouncePeers(req.InfoHash, req.Left == 0, int(req.NumWant), req.Peer)
+		resp.IPv4Peers, err = h.getPeers(req.InfoHash, req.Left == 0, int(req.NumWant), req.Peer)
 	case net.IPv6len:
-		resp.IPv6Peers, err = h.store.AnnouncePeers(req.InfoHash, req.Left == 0, int(req.NumWant), req.Peer)
+		resp.IPv6Peers, err = h.getPeers(req.InfoHash, req.Left == 0, int(req.NumWant), req.Peer)
 	default:
 		return ctx, ErrInvalidIP
 	}
 
 	return ctx, err
+}
+
+func (h *responseHook) getPeers(ih bittorrent.InfoHash, seeder bool, numWant int, p bittorrent.Peer) ([]bittorrent.Peer, error) {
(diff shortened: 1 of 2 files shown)


middleware/hooks.go
```diff
@@ -102,16 +102,31 @@ func (h *responseHook) HandleAnnounce(ctx context.Context, req *bittorrent.Annou
...
 	return ctx, err
 }
 
+func (h *responseHook) getPeers(ih bittorrent.InfoHash, seeder bool, numWant int, p bittorrent.Peer) ([]bittorrent.Peer, error) {
+	peers, err := h.store.AnnouncePeers(ih, seeder, numWant, p)
+	if err != nil && err != storage.ErrResourceDoesNotExist {
+		return nil, err
+	}
+
+	// Insert announcing peer.
```
jzelinskie: Can we expand this comment to include the fact that certain clients expect at a minimum to always at least see themselves reflected in their announces?

middleware/hooks.go
```diff
@@ -102,16 +102,31 @@ func (h *responseHook) HandleAnnounce(ctx context.Context, req *bittorrent.Annou
...
+	if len(peers) < numWant || peers == nil {
+		return append(peers, p), nil
+	}
+
+	peers[len(peers)-1] = p
```
jzelinskie: This could explode if a client asks for `numWant == 0` and then we return an `len(peers) == 0`, then we're attempting to index `-1`.

mrd0ll4r: Good point, thx

mrd0ll4r:
ptal

storage/storage.go
```diff
@@ -51,6 +51,8 @@ type PeerStore interface {
 	// - if seeder is true, should ideally return more leechers than seeders
 	// - if seeder is false, should ideally return more seeders than
 	//   leechers
+	//
+	// Returns ErrDNE if the provided infoHash is not tracked.
```
jzelinskie: let's use the full error name here instead of `ErrDNE`

mrd0ll4r:
ptal

PR #236, opened 2016-09-28
http: fix IPv4 addresses not 4 bytes long

### LINKED ISSUE #230: `net.ParseIP` parses IPv4 addresses to 16-byte `net.IP`s
This becomes a problem for the HTTP server (e.g. it uses the remote address of the HTTP request (or some other source for a string-representation of the remote's IP) and calls `net.ParseIP` on it).

This 16-byte IPv4-in-IPv6 address then moves through the logic and creates all kinds of problems.

mrd0ll4r:
Fixes #230

### DIFF
--- a/frontend/http/parser.go
+++ b/frontend/http/parser.go
@@ -79,6 +79,11 @@ func ParseAnnounce(r *http.Request, realIPHeader string, allowIPSpoofing bool) (
 		return nil, bittorrent.ClientError("failed to parse peer IP address")
 	}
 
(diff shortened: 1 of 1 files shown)


jzelinskie:
I'm not sure if this is the correct place to fix the bug. Maybe this should be done in the HTTP frontend only. [I don't think UDP needs it.](https://play.golang.org/p/lkl8RpWQq1)

mrd0ll4r:
Makes sense, ptal

PR #237, opened 2016-09-29
jwt: verifying JWT are now client errors

jzelinskie:
These were previously raising a 500 in the HTTP server; rather, they
should return 200 but explain the the client that their JWTs are
invalid.

### DIFF
--- a/tracker/jwt.go
+++ b/tracker/jwt.go
@@ -7,7 +7,6 @@ package tracker
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 	"net/http"
 	"net/url"
 	"time"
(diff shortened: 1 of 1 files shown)


tracker/jwt.go
```diff
@@ -7,14 +7,15 @@ package tracker
...
-	"fmt"
 	"net/http"
 	"net/url"
 	"time"
 
 	oidchttp "github.com/coreos/go-oidc/http"
 	"github.com/coreos/go-oidc/jose"
 	"github.com/golang/glog"
+
+	"github.com/chihaya/chihaya/tracker/models"
```
mrd0ll4r: This package does not seem right.

jzelinskie: did you notice which branch this is? it compiles.

mrd0ll4r: Oh no I completely overlooked that! Sorry about that...

PR #242, opened 2016-10-02
middleware: fix scrape data in announces

### LINKED ISSUE #241: First announce has weird complete/incomplete counters
If we announce this:
```
http://127.0.0.1:6881/announce?compact=1&downloaded=0&event=started&info_hash=%21%03%AE%92%00%81b%FB%F6x%AE%E1S5%C1%DE%D0%92%2C%27&ip=64.64.166.68&left=100&numwant=50&peer_id=-POKE64-000000017574&port=17574&uploaded=0
```

we get this:
```
d12:min intervali0e5:peers6:@@�DD�8:completei0e10:incompletei0e8:intervali0ee
```

Which contains the announcing peer, but also states that there are zero seeders and zero leechers.

I don't know how bad this is in practice, but we should probably make the response hook smart enough to also increment the counters for an `ErrResourceDoesNotExist`, the same way we insert the announcing peer.

jzelinskie:
This change adjusts it such that only empty Swarms return the request
Peer. This also increments the Scrape data properly if a peer is added.

Fixes #241.

### DIFF
--- a/middleware/hooks.go
+++ b/middleware/hooks.go
@@ -96,36 +96,43 @@ func (h *responseHook) HandleAnnounce(ctx context.Context, req *bittorrent.Annou
 		return ctx, nil
 	}
 
+	// Add the Scrape data to the response.
 	s := h.store.ScrapeSwarm(req.InfoHash, len(req.IP) == net.IPv6len)
 	resp.Incomplete = s.Incomplete
 	resp.Complete = s.Complete
 
-	switch len(req.IP) {
-	case net.IPv4len:
-		resp.IPv4Peers, err = h.getPeers(req.InfoHash, req.Left == 0, int(req.NumWant), req.Peer)
-	case net.IPv6len:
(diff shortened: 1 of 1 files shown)


mrd0ll4r:
The code LGTM, but:

1. Is it true that these clients only expect themselves in an otherwise empty announce response?
2. Maybe we should generally increment the scrape data the way you did. Imagine this scenario: There is a swarm with one leecher. A seeder announces and receives a response with one peer and scrape data that indicates "there is one leecher and no seeder". Do clients expect the scrape data to look like it's up-to-date and valid?

Thoughts?

jzelinskie:
1. To my understanding this is true. This is how chihaya v1 works.
2. The reason why we don't do this generically is because we cannot know if the peer is already in the swarm or not. We could make a reasonable guess: if the request has the event value "started" or "stopped" they aren't in the swarm. By guessing, this data would probably be wrong with frequent packet loss or restarting a tracker with a bunch of active clients.

mrd0ll4r:
I'll open an issue to discuss this further, PR looks good!

PR #244, opened 2016-10-04
udp: fix IP bytes not being copied out of the packet

### LINKED ISSUE #243: UDP frontend erases Peer's IP from announce request
In [udp/frontend.go:214](https://github.com/chihaya/chihaya/blob/master/frontend/udp/frontend.go#L214) we call `t.logic.AfterAnnounce` in a separate goroutine. This causes `req.Peer.IP` to be set to zero (at least on my machine).

Find an annotated version of the code here: https://gist.github.com/mrd0ll4r/5ab421a8f0fef019f4f06a632cfac4e7
And a log produced by that: https://gist.github.com/mrd0ll4r/ec4ba9c75e0a8a0a8db665eadca0610c

mrd0ll4r:
Fixes #243

### DIFF
--- a/frontend/udp/frontend.go
+++ b/frontend/udp/frontend.go
@@ -128,10 +128,15 @@ func (t *Frontend) ListenAndServe() error {
 			defer t.wg.Done()
 			defer pool.Put(buffer)
 
+			if ip := addr.IP.To4(); ip != nil {
+				addr.IP = ip
+			}
+
 			// Handle the request.
 			start := time.Now()
 			action, err := t.handleRequest(
-				Request{buffer[:n], addr.IP},
+				// Make sure the IP is copied, not referenced.
+				Request{buffer[:n], append([]byte{}, addr.IP...)},
 				ResponseWriter{t.socket, addr},
 			)
 			recordResponseDuration(action, err, time.Since(start))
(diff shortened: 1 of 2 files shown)


jzelinskie:
Why not use the builtin `copy`?

mrd0ll4r:
Good point, I found and squashed another bug in the process :)

frontend/udp/frontend.go
```diff
@@ -128,10 +128,15 @@ func (t *Frontend) ListenAndServe() error {
...
+			if ip := addr.IP.To4(); ip != nil {
+				addr.IP = ip
+			}
+
 			// Handle the request.
 			start := time.Now()
 			action, err := t.handleRequest(
-				Request{buffer[:n], addr.IP},
+				// Make sure the IP is copied, not referenced.
+				Request{buffer[:n], append([]byte{}, addr.IP...)},
```
jzelinskie: why use append over `[]byte(addr.IP)`, is there some allocation being saved or something?

mrd0ll4r: Apparently `[]byte(addr.IP)` does not make a copy of the underlying array, but we need it to be a copy.

If it's not a copy, the changes introduced in parser.go will overwrite the bytes of `addr.IP` which will cause the response packet to be sent to whatever address the client spoofed.

See this link on the playground: https://play.golang.org/p/0NdtJU_jqX

jzelinskie: cool!

PR #245, opened 2016-10-12
Fix configure file parse

hustcat:
Fix configure file parse

### DIFF
--- a/cmd/chihaya/config.go
+++ b/cmd/chihaya/config.go
@@ -23,13 +23,13 @@ type hookConfig struct {
 // ConfigFile represents a namespaced YAML configation file.
 type ConfigFile struct {
 	MainConfigBlock struct {
-		middleware.Config
-		PrometheusAddr string              `yaml:"prometheus_addr"`
-		HTTPConfig     httpfrontend.Config `yaml:"http"`
-		UDPConfig      udpfrontend.Config  `yaml:"udp"`
-		Storage        memory.Config       `yaml:"storage"`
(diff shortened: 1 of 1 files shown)


mrd0ll4r:
I noticed this didn't parse correctly too.

However, I'd like to keep the config file as is, i.e. not have the `middleware` as a block just to hold the `announce_interval`.

There is an `inline` annotation, i.e.

```go
type A struct {
    B `yaml:",inline"`
    D int `yaml:"d"`
}

type B struct {
    C int `yaml:"c"`
}
```

cmd/chihaya/config.go
```diff
@@ -23,13 +23,13 @@ type hookConfig struct {
...
 type ConfigFile struct {
 	MainConfigBlock struct {
-		middleware.Config
-		PrometheusAddr string              `yaml:"prometheus_addr"`
-		HTTPConfig     httpfrontend.Config `yaml:"http"`
-		UDPConfig      udpfrontend.Config  `yaml:"udp"`
-		Storage        memory.Config       `yaml:"storage"`
-		PreHooks       []hookConfig        `yaml:"prehooks"`
-		PostHooks      []hookConfig        `yaml:"posthooks"`
+		middleware.Config `yaml:"middleware"`
```
mrd0ll4r: Change this to use the `inline` annotation, please.

hustcat:
@mrd0ll4r Thank you. I have changed it:)

mrd0ll4r:
LGTM, thanks for the PR! :)

PR #248, opened 2016-10-26
Fix JWT Middleware

jzelinskie:


mrd0ll4r:
Nice catch! This looks fine as it is, but what I'd love to have is a test. Preferably one that actually starts a server that provides a JWK set, creates a signed announce request and checks that with the middleware. For me the whole JWK stuff sounds good, and I understand how our end works, but for the average user it could be useful to see how the other components (key server, torrent signing) work/need to be set up.

Overkill? Maybe belongs in a README rather than a test?

jzelinskie:
Yeah, I still need to write docs for this middleware. I think I can get away with just writing unit tests to test all the behavior here.

mrd0ll4r:
Alright, this PR on it's own LGTM then :)

PR #249, opened 2016-10-26
add docs directory

jzelinskie:
i bought us a domain too!

Should we make this another repository?

### DIFF
--- a/docs/CNAME
+++ b/docs/CNAME
@@ -1,0 +1,1 @@
+chihaya.io


mrd0ll4r:
Nice job with the domain! And even a cool `.io` one :)

You mean like the typical `xxx.chihaya.io` repository? What do you plan to put in the `docs/` directory? Just markdown files?

My opinion is something like this: If we just have markdown files, they should go in the `docs/` directory in this repository. Whoever clones it also has the full documentation right there. If it's gonna be HTML/CSS or whatever else, open a new repo, I always hate finding HTML documentation in a project folder... that's my two cents.

jzelinskie:
OK. I vote we make a hugo static site on another repo and put all our markdown docs into the docs directory on chihaya, then we can just use a git submodule and a symlink.

PR #253, opened 2016-11-07
prand: add Container

mrd0ll4r:
I need this for the deniability and varinterval middleware to be able to handle multiple requests in parallel.

### DIFF
--- a/pkg/prand/prand.go
+++ b/pkg/prand/prand.go
@@ -1,0 +1,75 @@
+// Package prand allows parallel access to randomness based on indices or
+// infohashes.
+package prand
+
+import (
+	"encoding/binary"
+	"math/rand"
+	"sync"
+	"time"
+
+	"github.com/chihaya/chihaya/bittorrent"
+)
+
+type lockableRand struct {
+	*rand.Rand
+	*sync.Mutex
+}
+
+// Container is a container for sources of random numbers that can be locked
+// individually.
+type Container struct {
+	rands []lockableRand
+}
+
+// NewSeeded returns a new Container with num sources that are seeeded with
+// seed.
+func NewSeeded(num int, seed int64) *Container {
+	toReturn := Container{
+		rands: make([]lockableRand, num),
+	}
+
+	for i := 0; i < num; i++ {
+		toReturn.rands[i].Rand = rand.New(rand.NewSource(seed))
+		toReturn.rands[i].Mutex = &sync.Mutex{}
+	}
+
+	return &toReturn
+}
+
+// New returns a new Container with num sources that are seeded with the current
+// time.
+func New(num int) *Container {
+	return NewSeeded(num, time.Now().UnixNano())
+}
+
+// Get locks and returns the nth source.
+//
+// Get panics if n is not a valid index for this Container.
+func (s *Container) Get(n int) *rand.Rand {
+	r := s.rands[n]
+	r.Lock()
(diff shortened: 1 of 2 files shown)


jzelinskie:
Can we put this in a `pkg/prand` and add a package description that this allows parallel access to randomness.

mrd0ll4r:
ptal

jzelinskie:
I'm a fan of hiding locks from APIs. I'd rather have the locking behavior internal to the type so that the caller can't shoot themselves in the foot. Also coupling the locks with the thing they lock into the same object is another thing I usually do. So in this example `Container` could be a `[]struct{rand, mutex}` (pseudo-code).

mrd0ll4r:
Putting the locks with the rands into a struct should generate a more local layout in memory, makes sense, will do.

But I can't hide the locking from the user. Sure, I could call the methods `Get(n int)` or `GetByInfohash(ih)`, but the user still has to call some method to get a `*rand.Rand` and some method to return it.

mrd0ll4r:
I updated this btw, ptal :)

jzelinskie:
Let me just confirm I understand this PR:

You need randomness for the aforementioned two middleware. `math/rand`'s exported functions are already locked, but to enable performance you want multiple sources of randomness so that everything isn't contending for the same lock.

Can you post an example of where you'd use this package? Without seeing more, I can't understand why you wouldn't want to implement roughly the same API as [`math/rand.Rand`](https://godoc.org/math/rand#Rand).

mrd0ll4r:
> You need randomness for the aforementioned two middleware. math/rand's exported functions are already locked, but to enable performance you want multiple sources of randomness so that everything isn't contending for the same lock.

Yup, that's exactly it. If we have one of these middleware in the chain and they effectively run single-threaded our whole chain is stalled.

For example: The deniability middleware uses it [like this](https://github.com/mrd0ll4r/chihaya/blob/71edd693308c83836d767f533da297f218994d6e/middleware/deniability/deniability.go#L149). I implemented this PR manually a few lines earlier. I need basically the same functionality for the varinterval middleware, but don't want to copy the implementation.

Now for why this API: Implementing `math/rand.Rand` is huge and needs to behave (i.e. produce the same kind of numbers as) the stdlib. I thought about implementing a `rand.Source` that works in parallel, but in order to use that in parallel you'd need multiple `rand.Rand`s feeding off that, which brings us back to the multiple-`rand.Rand`s approach I took here.

A lot of benchmarking went into this: I tried channel-based solutions, lock-free slices of channels, and also pooling the actual values (let's say a `chan float64` that is written to by multiple goroutines calling `Float64()` on a local `rand.Rand`), but all of them perform poorly. The only thing that is as fast as (and occasionally faster than) this solution is a `sync.Pool`. That would be the only viable alternative, but I dislike it because noone really knows what it does and it introduces some nondeterminism. (Also the performance is very close, on my laptop the difference is ~4 ns.)

So yeah, implementing just the API I need (some `float64`s, `int`s, and `byte`s, iirc) would work, but is too slow and too specific to be a `pkg/prand`. All middleware that needs randomness needs to be able to handle multiple requests in parallel - I don't want to force future middleware to use a small subset of API `rand.Rand` provides.

jzelinskie:
This sounds fair to me. I'm willing to LGTM after a rebase.

mrd0ll4r:
Rebased, small optimization, ptal

PR #257, opened 2016-11-27
cmd/chihaya: add config reloading via SIGUSR1

### LINKED ISSUE #215: support reloading config via unix signal
Given a SIGHUP, I think chihaya should shutdown everything except for the peer store, and then start back up with a newly parsed config.

mrd0ll4r:
I made this a SIGUSR1 instead of a SIGHUP. I know I know, many things use SIGHUP for this kind of stuff, but I think that's because there was nothing else when they were written. We have user-defined signals now, so why not use them?

Fixes #215

### DIFF
--- a/cmd/chihaya/main.go
+++ b/cmd/chihaya/main.go
@@ -15,6 +15,7 @@ import (
 	httpfrontend "github.com/chihaya/chihaya/frontend/http"
 	udpfrontend "github.com/chihaya/chihaya/frontend/udp"
 	"github.com/chihaya/chihaya/middleware"
+	"github.com/chihaya/chihaya/storage"
 	"github.com/chihaya/chihaya/storage/memory"
 )
 
@@ -53,7 +54,6 @@ func rootCmdRun(cmd *cobra.Command, args []string) error {
 		}
 	}()
 
-	// Force the compiler to enforce memory against the storage interface.
 	peerStore, err := memory.New(cfg.Storage)
 	if err != nil {
 		return errors.New("failed to create memory storage: " + err.Error())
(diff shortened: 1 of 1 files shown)


cmd/chihaya/main.go
```diff
@@ -65,67 +64,89 @@ func rootCmdRun(cmd *cobra.Command, args []string) error {
...
+					log.Debug("Successfully restarted")
+					continue
+				}
+			case <-shutdown:
+			}
 
-		if udpFrontend != nil {
-			udpFrontend.Stop()
-		}
+			// Stop frontends
```
jzelinskie: After you reload the config, won't this code still execute and shutdown everything?

mrd0ll4r: That's what all the `continue`s are for - the jump back to the `for`, blocking on the `select` again. (I honestly wasn't sure whether `continue` in a `select` did some other funky stuff, but I tested it and it works :) )

jzelinskie: Why not explicitly put that code in the shutdown case?

mrd0ll4r: We also shut down gracefully on `SIGINT` and `SIGTERM`, i.e. on receiving from the `sig` channel, skipping the `if`, then falling out of the `select`

jzelinskie: I see now. Would it be more clear if we had two signal channels? One dedicated to config reloading and the other to handling termination? Then we can be explicit about the fallthrough to the shutdown logic.

mrd0ll4r: Good idea! ptal

cmd/chihaya/main.go
```diff
@@ -65,67 +64,92 @@ func rootCmdRun(cmd *cobra.Command, args []string) error {
...
-	}
-
-	if cfg.UDPConfig.Addr != "" {
-		udpFrontend = udpfrontend.NewFrontend(logic, cfg.UDPConfig)
 
-		go func() {
-			log.Infoln("started serving UDP on", cfg.UDPConfig.Addr)
-			if err := udpFrontend.ListenAndServe(); err != nil {
-				errChan <- errors.New("failed to cleanly shutdown UDP frontend: " + err.Error())
+			// Stop frontends
```
jzelinskie: I think moving this into the `case` block for _shutdown_ and then explicitly adding `fallthrough` inside the _quit_ case makes the logic more clear.

mrd0ll4r: You cannot `fallthrough` in `select`s: https://play.golang.org/p/TWkkBrxZap. You cannot even `goto` the other case. The clearest solution would be something like this (see commit), but meh. Any suggestions?

PR #259, opened 2016-11-28
initial helm package

jzelinskie:
This makes it possible for you to deploy Chihaya on Kubernetes in 2 commands:

```sh
$ helm package contrib/helm/chihaya
$ helm install chihaya-0.1.0.tgz
```

### DIFF
--- a/contrib/helm/chihaya/templates/service.yaml
+++ b/contrib/helm/chihaya/templates/service.yaml
@@ -1,0 +1,27 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "fullname" . }}
+  labels:
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+  annotations:
+    prometheus.io/scrape: "true"
+    prometheus.io/path: "/"
+    prometheus.io/port: {{ $v := .Values.config.chihaya.prometheus_addr | split ":" }}{{ $v._1 | quote }}
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+  - name: bittorrent-http
+    port: {{ $v := .Values.config.chihaya.http.addr | split ":" }}{{ $v._1 }}
+    targetPort: {{ $v := .Values.config.chihaya.http.addr | split ":" }}{{ $v._1 }}
+    protocol: TCP
+  - name: bittorrent-udp
+    port: {{ $v := .Values.config.chihaya.udp.addr | split ":" }}{{ $v._1 }}
+    targetPort: {{ $v := .Values.config.chihaya.udp.addr | split ":" }}{{ $v._1 }}
+    protocol: UDP
+  - name: metrics
+    port: {{ $v := .Values.config.chihaya.prometheus_addr | split ":" }}{{ $v._1 }}
+    targetPort: {{ $v := .Values.config.chihaya.prometheus_addr | split ":" }}{{ $v._1 }}
+    protocol: TCP
+  selector:
+    app: {{ template "fullname" . }}
--- a/contrib/helm/chihaya/values.yaml
+++ b/contrib/helm/chihaya/values.yaml
@@ -1,0 +1,33 @@
+replicaCount: 1
+image:
+  repository: quay.io/jzelinskie/chihaya
+  tag: v2.0.0-rc.1
+  pullPolicy: IfNotPresent
+service:
+  name: chihaya
+  type: ClusterIP
+resources:
+  limits:
+    cpu: 100m
+    memory: 1Gi
+config:
+  chihaya:
+    announce_interval: 15m
+    prometheus_addr: 0.0.0.0:6880
+    http:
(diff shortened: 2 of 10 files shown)


contrib/helm/chihaya/values.yaml
```diff
@@ -0,0 +1,33 @@
...
+      allow_ip_spoofing: true
+      real_ip_header: X-Forwarded-For
+      read_timeout: 5s
+      write_timeout: 5s
+      request_timeout: 5s
+    udp:
+      addr: 0.0.0.0:6881
+      allow_ip_spoofing: true
+      max_clock_skew: 10s
+      private_key: "/K8AybyuKkob03fZJb+3Fu87eVcBNWsvVyDDxGd3wfvVGix10j6iGDEyS/ELPvQIIZvUPl2PnyfnpG6g1YJ/sw=="
```
jzelinskie: I think I should probably leave this blank and change chihaya to generate something when this key isn't present.

mrd0ll4r:
I have no way of testing this, so no comment on that. Just change the private key, LGTM otherwise.

frontend/udp/frontend.go
```diff
@@ -66,6 +67,16 @@ type Frontend struct {
 
 // NewFrontend allocates a new instance of a Frontend.
 func NewFrontend(logic frontend.TrackerLogic, cfg Config) *Frontend {
+	if cfg.PrivateKey == "" {
```
mrd0ll4r: Can you add a log message for this? The user should know that this makes old connection IDs invalid.

frontend/udp/frontend.go
```diff
@@ -66,6 +68,20 @@ type Frontend struct {
...
+		log.Warning("no UDP private key provided: one will be generated for this instance")
+
+		buf := make([]byte, 2048)
+		_, err := rand.Read(buf)
+		if err != nil {
+			panic("system failed to provide sufficient randomness: " + err.Error())
+		}
+		cfg.PrivateKey = string(buf)
+
+		log.Info("the following UDP private key was generated: %s", cfg.PrivateKey)
```
mrd0ll4r: Dumping random bytes to a log probably creates funky side-effects, let's not do that

jzelinskie: just gotta change it to %x ;)

jzelinskie: I realized that if I do, then we'd have to parse this field as hex, too. Thoughts?

mrd0ll4r: Oh right, that makes sense. Or specify a file to read the key from? But I guess that doesn't work well with docker?

jzelinskie: So this has blocked this PR for a bit. Should we enforce that this is a string rather than []byte? There isn't a crazy requirement around cryptographic security for this value.

mrd0ll4r: Yeah, let's make it a string. The deniability PR has a function to generate a random string, maybe we should pull that out and put it somewhere in `pkg/` for re-use.

jzelinskie:
I updated this to make it a little more sane. Let's merge it and then if/when we get it merged upstream in kubernetes/charts incubator, we can delete it.

contrib/helm/chihaya/templates/service.yaml
```diff
@@ -0,0 +1,27 @@
...
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+  annotations:
+    prometheus.io/scrape: "true"
+    prometheus.io/path: "/"
+    prometheus.io/port: {{ $v := .Values.config.chihaya.prometheus_addr | split ":" }}{{ $v._1 | quote }}
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+  - name: bittorrent-http
+    port: {{ $v := .Values.config.chihaya.http.addr | split ":" }}{{ $v._1 }}
```
mrd0ll4r: How does this splitting work with IPv6?

contrib/helm/chihaya/values.yaml
```diff
@@ -0,0 +1,32 @@
...
+      request_timeout: 5s
+    udp:
+      addr: 0.0.0.0:6881
+      allow_ip_spoofing: true
+      max_clock_skew: 10s
+    storage:
+      gc_interval: 14m
+      peer_lifetime: 15m
+      shards: 16
+      max_numwant: 50
```
mrd0ll4r: This is not a parameter of `storage` anymore - do we have to keep the two configs synchronized by hand?

jzelinskie: Yea, we'll have to. Maybe we should wait for a 2.0 cut before we suggest it upstream. I'll try and get this running tonight with HEAD.

contrib/helm/chihaya/templates/service.yaml
jzelinskie: I don't think Kubernetes default overlay network (Flannel) supports IPv6.

PR #260, opened 2016-12-12
stopper: move to pkg package

jzelinskie:
This also adds a package docs for the stopper package.


jzelinskie:
I agree that stuttering sucks, but let's address that in a different PR.

PR #270, opened 2017-01-22
docs: add github redirect

jzelinskie:


mrd0ll4r:
Is this the right place? My understanding was to put that file here: https://github.com/chihaya/chihaya.github.io
So it redirects users to the repository? And then we put the markdown documentation into `docs/` ?

jzelinskie:
We can get rid of those repositories and put everything here. By exposing our this repository's `docs/` as GH pages, and pointing the CNAME to that. No need to keep around an extra repository.

PR #275, opened 2017-01-23
frontend/udp: rm copypasta metric recording

### LINKED ISSUE #271: Prometheus has action=action metric floating around
This ([frontend/udp/frontend.go:23](https://github.com/chihaya/chihaya/blob/master/frontend/udp/frontend.go#L23)) line causes prometheus to have one entry with `action=action` floating around. This is somewhat annoying, because it also creates all the quantile buckets (?), which are never used again. Same applies to HTTP. 

Can we get rid of that?

jzelinskie:
Fixes #271.

### DIFF
--- a/frontend/http/frontend.go
+++ b/frontend/http/frontend.go
@@ -19,7 +19,6 @@ import (
 
 func init() {
 	prometheus.MustRegister(promResponseDurationMilliseconds)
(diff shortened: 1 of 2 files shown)


mrd0ll4r:
This also applies to HTTP, can you remove that line too, please?

PR #276, opened 2017-01-23
example_config: add initial documentation

jzelinskie:
I left the middleware undocumented because I'm not sure what we want to do there.

### DIFF
--- a/example_config.yaml
+++ b/example_config.yaml
@@ -1,29 +1,72 @@
 chihaya:
+  # The interval communicated with BitTorrent clients informing them how
+  # frequently they should announce in between client events.
   announce_interval: 15m
-  prometheus_addr: localhost:6880
+
+  # The network interface that will bind to an HTTP endpoint that can be
+  # scraped by an instance of the Prometheus time series database.
+  # For more info see: https://prometheus.io
+  prometheus_addr: 0.0.0.0:6880
+
+  # The maximum number of peers returned in an announce.
   max_numwant: 50
+
+  # The default number of peers returned in an announce.
   default_numwant: 25
 
+  # This block defines configuration for the tracker's HTTP interface.
+  # If you do not wish to run this, delete this section.
   http:
+    # The network interface that will bind to an HTTP server for serving
+    # BitTorrent traffic.
     addr: 0.0.0.0:6881
+
+    # When enabled, the IP address used to connect to the tracker overrides
+    # the value clients advertise.
     allow_ip_spoofing: false
+
+    # The HTTP Header containing the IP address of the client.
+    # This is only necessary if using a reverse proxy.
     real_ip_header: x-real-ip
+
+    # The timeout durations for HTTP requests.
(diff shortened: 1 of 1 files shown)


example_config.yaml
```diff
@@ -1,49 +1,89 @@
...
+
+  # The network interface that will bind to an HTTP endpoint that can be
+  # scraped by an instance of the Prometheus time series database.
+  # For more info see: https://prometheus.io
+  prometheus_addr: 0.0.0.0:6880
+
+  # The maximum number of peers returned in an announce.
   max_numwant: 50
+
+  # The default number of peers returned in an announce.
```
mrd0ll4r: ...if the client provides an invalid `num_want` parameter, i.e. `<=0`

example_config.yaml
```diff
@@ -1,49 +1,89 @@
...
   default_numwant: 25
 
+  # This block defines configuration for the tracker's HTTP interface.
+  # If you do not wish to run this, delete this section.
   http:
+    # The network interface that will bind to an HTTP server for serving
+    # BitTorrent traffic.
```
mrd0ll4r: We should probably make a comment about how to specify IPv6 addresses: In quotation marks, using square brackets.

example_config.yaml
```diff
@@ -1,49 +1,89 @@
...
     addr: 0.0.0.0:6881
+
+    # When enabled, the IP address used to connect to the tracker overrides
+    # the value clients advertise.
```
mrd0ll4r: It's the other way around, isn't it?

example_config.yaml
```diff
@@ -1,49 +1,89 @@
...
     allow_ip_spoofing: false
+
+    # The HTTP Header containing the IP address of the client.
+    # This is only necessary if using a reverse proxy.
```
mrd0ll4r: Also: Does this only work with `allow_ip_spoofing=true`?

jzelinskie: IIRC, if the client doesn't provide an IP we fallback to the RemoteAddr as a last resort.

example_config.yaml
jzelinskie: yep, I wrote this wrong.

example_config.yaml
jzelinskie: Yeah, that's an unfortunate side effect of YAML

example_config.yaml
jzelinskie: Unrelated: is numwant = 0 actually a bad value?

mrd0ll4r: Oh, numwant is optional, hmm. You mean something like seeders actually announcing with numwant=0? Problem is: How do we distinguish that from `numwant` missing (=default)? Maybe we should make `numwant` a signed integer again, and set that to -1?

jzelinskie: As discussed on IRC, we could make the sanitization a function of the bittorrent package along with a NumWant type that has semantic meaning for the -1 value.

PR #279, opened 2017-01-25
udp: make connection ID handling faster

mrd0ll4r:
This adds benchmarks for connection ID generation and verification. Also, the assembly implementation of sha256 improves performance for both by about 500ns (from ~2700ns to ~2200ns).

Click [here](https://github.com/minio/sha256-simd) for the sha256-simd repo.

### DIFF
--- a/frontend/udp/connection_id.go
+++ b/frontend/udp/connection_id.go
@@ -2,10 +2,11 @@ package udp
 
 import (
 	"crypto/hmac"
-	"crypto/sha256"
 	"encoding/binary"
 	"net"
 	"time"
+
+	sha256 "github.com/minio/sha256-simd"
 )
 
 // ttl is the number of seconds a connection ID should be valid according to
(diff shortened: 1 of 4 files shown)


frontend/udp/connection_id.go
```diff
@@ -2,10 +2,11 @@ package udp
 
 import (
 	"crypto/hmac"
-	"crypto/sha256"
 	"encoding/binary"
 	"net"
 	"time"
+
+	sha256 "github.com/minio/sha256-simd"
```
jzelinskie: update glide to include this dep

mrd0ll4r:
glide updated, ptal

jzelinskie:
Just want to absolutely make sure this fails over to something if SIMD isn't available before I LGTM

mrd0ll4r:
It does, see [the New func](https://github.com/minio/sha256-simd/blob/master/sha256.go#L86) :)

PR #282, opened 2017-01-29
udp: fix ordering of scrapes

### LINKED ISSUE #280: UDP Scrape Responses are not in order
[BEP 15](http://bittorrent.org/beps/bep_0015.html) specifies the UDP tracker protocol (now with IPv6!). A scrape response contains a list of triples `<seeders, completed, leechers>`, without referencing the infohash these values belong to. They are matched in order with the infohashes requested in the scrape request.

We are currently iterating over a map to serialize the scrape response, we must change this to keep the `Scrape`s in the order they were requested.

mrd0ll4r:
Fixes #280.

Or maybe we should change the `ScrapeResponse` to contain a slice of `Scrape`s and make them contain an `InfoHash`?

### DIFF
--- a/bittorrent/bittorrent.go
+++ b/bittorrent/bittorrent.go
@@ -100,12 +100,16 @@ type ScrapeRequest struct {
 }
 
 // ScrapeResponse represents the parameters used to create a scrape response.
+//
+// The Scrapes must be in the same order as the InfoHashes in the corresponding
+// ScrapeRequest.
 type ScrapeResponse struct {
-	Files map[InfoHash]Scrape
+	Files []Scrape
 }
 
 // Scrape represents the state of a swarm that is returned in a scrape response.
 type Scrape struct {
+	InfoHash   InfoHash
 	Snatches   uint32
 	Complete   uint32
 	Incomplete uint32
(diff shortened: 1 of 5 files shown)


frontend/udp/writer.go
```diff
@@ -55,19 +59,26 @@ func WriteAnnounce(w io.Writer, txID []byte, resp *bittorrent.AnnounceResponse,
...
 	buf := newBuffer()
 
 	writeHeader(buf, txID, scrapeActionID)
 
-	for _, scrape := range resp.Files {
+	// Preserve the ordering.
+	for _, ih := range infohashes {
+		scrape, ok := resp.Files[ih]
+		if !ok {
+			buf.free()
```
jzelinskie: can we just `defer buf.free()` after the call to `newBuffer` to avoid future mistakes?

mrd0ll4r: But `defer`s are slow :(

jzelinskie: yea, we had a bug a few years ago because we forgot to release a mutex one place.
i wish defer had slightly different semantics and was basically just a macro. however, i acknowledge it's cost and we should probably avoid it.

mrd0ll4r:
Alright, I changed the type. It's a little worse than expected, this bubbled up into the storage now, too.

jzelinskie:
This feels like something that needs a unit test, too.

PR #284, opened 2017-02-01
prometheus: record IP AddressFamily

jzelinskie:


mrd0ll4r:
I thought about this some more and don't think we need it for the frontends. Each frontend only listens on one address, which is either IPv4 or IPv6, so all traffic it gets will be either IPv4 or IPv6.

It could be useful for the storage, because you _could_ run HTTP on one address family and UDP on the other, however much sense that might make.

jzelinskie:
>Each frontend only listens on one address

If you listen on 0.0.0.0, you bind to all interfaces.

bittorrent/bittorrent.go
```diff
@@ -120,7 +120,8 @@ type AddressFamily uint8
 
 // AddressFamily constants.
 const (
-	IPv4 AddressFamily = iota
+	Unknown AddressFamily = iota
```
mrd0ll4r: Let's not add a third type to this boolean enum.

frontend/http/frontend.go
```diff
@@ -30,19 +30,27 @@ var promResponseDurationMilliseconds = prometheus.NewHistogramVec(
...
 		Buckets: prometheus.ExponentialBuckets(9.375, 2, 10),
 	},
-	[]string{"action", "error"},
+	[]string{"action", "address_family", "error"},
 )
 
 // recordResponseDuration records the duration of time to respond to a Request
 // in milliseconds .
-func recordResponseDuration(action string, err error, duration time.Duration) {
+func recordResponseDuration(action string, af bittorrent.AddressFamily, err error, duration time.Duration) {
```
mrd0ll4r: You could make this a pointer to an `AddressFamily` and pass `nil` to indicate an unknown AF.

jzelinskie: The pointer's bigger than the af. I think we should use 0x0 to be unknown.

PR #295, opened 2017-02-15
frontend/*: only record ClientErrors to Prometheus

### LINKED ISSUE #294: Ensure errors are static to avoid creating streams in Prometheus
Weird title, problem is this: If URL escaping throws an error, we return something like "invalid URL escape `something`", where `something` is what caused the error. We pass those errors to prometheus, which can generate a lot of entries in the error dimension.

jzelinskie:
All ClientErrors are constant and should not cause Prometheus streams to
be generated for all possible failure scenarios in the program.

Fixes #294.

### DIFF
--- a/frontend/http/frontend.go
+++ b/frontend/http/frontend.go
@@ -39,7 +39,11 @@ var promResponseDurationMilliseconds = prometheus.NewHistogramVec(
 func recordResponseDuration(action string, af *bittorrent.AddressFamily, err error, duration time.Duration) {
 	var errString string
 	if err != nil {
-		errString = err.Error()
+		if _, ok := err.(bittorrent.ClientError); ok {
+			errString = err.Error()
+		} else {
+			errString = "internal error"
+		}
 	}
 
 	var afString string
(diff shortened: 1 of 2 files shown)


mrd0ll4r:
This makes non-`ClientError` failures look like successful requests, doesn't it? Because `error` remains empty?

jzelinskie:
If you define success by lack of error dimension in the metric, yes.
I could assign the errString to "internal error" for non-ClientErrors.

jzelinskie:
updated

frontend/http/frontend.go
```diff
@@ -39,7 +39,11 @@ var promResponseDurationMilliseconds = prometheus.NewHistogramVec(
 func recordResponseDuration(action string, af *bittorrent.AddressFamily, err error, duration time.Duration) {
 	var errString string
```
mrd0ll4r: Do you think it'd be slower to change this to `errString := "internal error"` and drop the `else` case?

jzelinskie: If there's no error, it should be empty

mrd0ll4r: Uh, yeah...mb.

mrd0ll4r:
We should probably do a writeup of how frontends should work, including stuff like this.

jzelinskie:
I think this stuff should likely be documented in a document dedicated metrics, but I agree. We should sit down and write docs.

PR #303, opened 2017-03-05
Reuse port for HTTP & HTTPS

jzelinskie:
Thanks for introducing the cmux library to me @cooperlees.

### DIFF
--- a/frontend/http/frontend.go
+++ b/frontend/http/frontend.go
@@ -7,11 +7,13 @@ import (
 	"crypto/tls"
 	"net"
 	"net/http"
+	"strings"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/julienschmidt/httprouter"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/soheilhy/cmux"
 
 	"github.com/chihaya/chihaya/bittorrent"
 	"github.com/chihaya/chihaya/frontend"
@@ -73,7 +75,8 @@ type Config struct {
 
 // Frontend holds the state of an HTTP BitTorrent Frontend.
 type Frontend struct {
-	s *http.Server
+	http  *http.Server
+	https *http.Server
 
 	logic frontend.TrackerLogic
 	Config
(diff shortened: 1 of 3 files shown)


mrd0ll4r:
What's the impact on performance, and do we lose HTTP/2 support?

Also, what if the tracker specifically only wants HTTPS and no HTTP?

jzelinskie:
HTTP2 is only over TLS.

mrd0ll4r:
> I can't particularly think of a use case that requires this.
Private trackers like to do this. Yes, I know, no private tracker is using chihaya yet, but I do think people would one HTTPS only. We could make HTTP optional the same way HTTPS is now, if that makes sense?

jzelinskie:
This is a fair point. It fundamentally falls down to a configuration problem. Is there a succinct way to describe the following states: (HTTP_ONLY, HTTPS_ONLY, BOTH)?

We _could_ add another frontend to the config `https` and enforce in the main package that the TLS details are configured, then we could just initialize 3 frontends.

I don't know if I want to go totally down the rabbithole and let people create arbitrary numbers of frontends.

mrd0ll4r:
So people would set `Addr` for both the `http` and `https` frontends to the same value, right? I think that might confuse people :D

Maybe just add a bool `https_only` to the `http` blockt and error if no TLS config is given?

frontend/http/frontend.go
```diff
@@ -104,31 +113,13 @@ func (t *Frontend) handler() http.Handler {
...
-				panic("http: connection transitioned to unknown state")
-			}
-		},
+	l, err := net.Listen("tcp", t.Addr)
+	if err != nil {
+		return err
 	}
-	t.s.SetKeepAlivesEnabled(false)
+
+	mux := cmux.New(l)
```
psaab: I would also add
mux.SetReadTimeout(t.ReadTimeout)

I found this was needed to keep the number of goroutines and fds open to a normal levels.

mrd0ll4r: Out of interest, what client are you using?

cooperlees: 99.99% our own wrapped libtorrent-rasterbar clients. There is some legacy 0.16.16 (I think - could of been updated) and majority 1.0.8 or something around there floating around. Why do you ask?

jzelinskie:
I think I'm going to ultimate just run both servers and if you only want to run HTTPS, you can simply not expose the port (by using something like iptables or not forwarding the port if you're using Docker).

psaab:
The feature is working extremely well for us in production now (off the 1.0 branch).  It really can simplify the entire process if you just run everything on a single port.

frontend/http/frontend.go
mrd0ll4r: Because of the weird behaviour with reads apparently timing out for you. If I'm not mistaken, otherwise the `SetReadTimeout` would have an effect, right?

jzelinskie:
I think I'm going to close this for now.
We should open an issue around this so that we can discuss it more.

Issue #298, opened 2017-02-17
Document how frontends should behave


mrd0ll4r:
We should make a note somewhere about how frontends should behave.
This includes things like metrics, errors being returned by the logic, handling of those errors, (soon: request sanitization) etc.

EDIT: They should use the context returned from the PreHooks for the PostHook chain, to allow communication.

EDIT2: They use the correct context. So just documentation now.

mrd0ll4r:
Fixed in #305

PR #305, opened 2017-03-06
docs: add frontend documentation

mrd0ll4r:
Taking a shot at #298

### DIFF
--- a/docs/frontend.md
+++ b/docs/frontend.md
@@ -1,0 +1,108 @@
+# Frontends
+
+A _Frontend_ is a component of Chihaya that serves a BitTorrent tracker on one protocol.
+The frontend accepts, parses and sanitizes requests, passes them to the _Logic_ and writes responses to _Clients_.
+
+This documentation first gives a high-level overview of Frontends and later goes into implementation specifics.
+Users of Chihaya are expected to just read the first part - developers should read both.
+
+## Functionality
+
+A Frontend serves one protocol, for example HTTP ([BEP 3]) or UDP ([BEP 15]).
+It listens for requests and usually answers each of them with one response, a basic overview of the control flow is:
+
+1. Read the request.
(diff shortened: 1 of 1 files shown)


docs/frontend.md
```diff
@@ -0,0 +1,103 @@
+# Frontends
+
+A _frontend_ is a component of Chihaya that serves a BitTorrent tracker on one protocol.
+The frontend accepts, parses and sanitizes requests, passes them to the _backend_ (or _logic_) and writes responses to _Clients_.
```
jzelinskie: Let's not use the word backend because it doesn't exist anywhere in the code or docs.

The reality is we should decide whether the documentation is going to target users or developers.
If it's the former, we shouldn't bother talking about "Logic" because it's an interface that's implemented by our middleware. If it's the latter, we should decide what is the division between doc comments and mardown docs.

docs/frontend.md
```diff
@@ -0,0 +1,103 @@
...
+Frontends must not return errors that are not a `bittorrent.ClientError` to the Client.
+A message like `internal server error` should be used instead.
+
+#### Request Sanitization
+
+The `TrackerLogic` expects sanitized requests in order to function properly.
+The constraints given here are for performance reasons not currently checked at runtime.
+
+Note that there is a function in the `bittorrent` package that enforces these constraints.
+(TODO provide name and usage.)
```
jzelinskie: we should use the normal todo syntax

```
TODO(username): task
````

docs/frontend.md
mrd0ll4r: I thought about targeting both. The general public would be interested in the general stuff, like the functionality and the available frontends. Developers would, in addition to that, read the part about implementing a frontend.  Makes sense?

mrd0ll4r:
ptal

PR #307, opened 2017-04-04
http: fix HTTP shutdown panicking

mrd0ll4r:


frontend/http/frontend.go
```diff
@@ -144,7 +144,8 @@ func (t *Frontend) ListenAndServe() error {
 	}
 
 	// Start the HTTP server and gracefully handle any network errors.
-	if err := t.s.ListenAndServe(); err != nil {
+	if err := t.s.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+		// TODO(mrd0ll4r): Do we still need this?
```
jzelinskie: I don't think so. Let's just log the error here and not bother checking the net.OpError stuff. That was all specific to graceful

mrd0ll4r:
ptal. I'm returning the error now, which should cause it to bubble up and bring chihaya down. Thoughts?

mrd0ll4r:
ptal, also fixed some wording in main.go (neither the HTTP frontend nor the UDP frontend returns errors on un-clean shutdown anymore).

PR #308, opened 2017-04-20
Make numwant in http request optional

TheOriginalWinCat:
The default numwant should be used if either the provided numwant is zero or omitted.

### DIFF
--- a/frontend/http/parser.go
+++ b/frontend/http/parser.go
@@ -63,7 +63,7 @@ func ParseAnnounce(r *http.Request, realIPHeader string, allowIPSpoofing bool) (
 	}
 
 	numwant, err := qp.Uint64("numwant")
-	if err != nil {
+	if err != nil && err != bittorrent.ErrKeyNotFound {
 		return nil, bittorrent.ClientError("failed to parse parameter: numwant")
 	}
 	request.NumWant = uint32(numwant)


mrd0ll4r:
We had a discussion about this some time ago and came to the conclusion that the default should only be used when it's actually missing from the request. We figured that we need to distinguish this special case from all other values, we intend to do that in one swoop with changes to the request sanitation, iirc.

Nonetheless, I agree that for now the HTTP frontend should not error on missing numwant.

LGTM, @jzelinskie thoughts?

jzelinskie:
I agree with @mrd0ll4r. This change looks reasonable -- it just needs a fix to the commit message.

Thanks for the first-time PR! Did you have any problems learning the codebase enough to make this change?

TheOriginalWinCat:
@mrd0ll4r My motivation for this change was that rTorrent by default don't specify a numwant in the announce requests unless you explicitly set a value. The BitTorrent spec which states the trackers default should be used if either the numwant is zero or not provided. So it seamed like a sane thing to create a PR for to see if this was intended behavior or a bug.

Yes sorry, stupidly I only read the contributing guidelines after submitting PR and Id made the commit before I thought about making a PR. I will fix the commit message.

@jzelinskie I've not read / written any go since go1.3 so most of my time studying the codebase was spent refreshing my knowledge of go. 

The only problem I ran into is that if I have debug logging enabled and send a restart signal the restart code is run twice causing a error since it can't bind to same location etc, and I can't see why that would happen.

mrd0ll4r:
> @mrd0ll4r My motivation for this change was that rTorrent by default don't specify a numwant in the announce requests unless you explicitly set a value. The BitTorrent spec which states the trackers default should be used if either the numwant is zero or not provided. So it seamed like a sane thing to create a PR for to see if this was intended behavior or a bug.
You're totally right, and I didn't want to discourage you :) Thanks for the PR!

> The only problem I ran into is that if I have debug logging enabled and send a restart signal the restart code is run twice causing a error since it can't bind to same location etc, and I can't see what would cause it to run twice.
That's interesting, can you file an issue for that? Wanna look into it.

TheOriginalWinCat:
@mrd0ll4r  I have updated the commit message and filed an issue for what I described.

mrd0ll4r:
Thanks for the issue. Sorry to bother you about the commit message one more time, but: Can you make it something like `http: make numWant optional`? We try to have the impacted package at the beginning, to see what commit is doing what.
Apart from that, looks all good :+1:

PR #310, opened 2017-04-30
Refactor root run command

### LINKED ISSUE #309: Restart logic fails if debug logging is enabled
If I send a restart signal when debug logging is enabled it fails to restart as it tries to bind twice.

here is a log of such behavior: 

> level = debug msg = "debug logging enabled"
> level = info msg = "started serving HTTP on 0.0.0.0:6881"
> level = info msg = "started serving prometheus stats on 0.0.0.0:6880"
> level = info msg = "Got signal to restart"
> level = debug msg = "Stopping frontends"
> level = debug msg = "Stopping logic"
> level = debug msg = "Restarting logic"
> level = debug msg = "Restarting frontends"
> level = debug msg = "Successfully restarted"
> level = info msg = "started serving HTTP on 0.0.0.0:6881"
> level = debug msg = "Stopping frontends"
> level = debug msg = "Stopping logic"
> level = debug msg = "Stopping storage"
> level = debug msg = "Stopping frontends"
> level = debug msg = "Stopping logic"
> level = debug msg = "Stopping storage"
> level = fatal msg = "http: failed to run HTTP server: listen tcp 0.0.0.0:6881: bind: address already in use"

So from the looks of things it seams to be running the quit or shutdown logic also.
In a attempt to confirm this I added logging to the quit and shutdown statements like so:

`

			case <-quit:

				log.Debug("quit signal")

				stop(udpFrontend, httpFrontend, logic, errChan, peerStore)

			case <-shutdown:

				log.Debug("shutdown signal")

				stop(udpFrontend, httpFrontend, logic, errChan, peerStore)
`

After adding the additional debug lines, it now behaves correctly. odd.
I then tried removing the two debug lines I added above and proceeded to remove the last debug log call in the restart case like so:

`//	log.Debug("Successfully restarted")`

And this also resolved the issue.
So it seams something about there been a call to log.Debug at the end of the case for restart that is causing other cases to run?

I'm using go1.8.1

jzelinskie:
This change refactors a bunch of the state of execution into its own
object. It also attempts to simplify stopping and adjusts some other
packages to integrate with the stopper interface.

I think it partially fixes #309.

If you build this and spam SIGUSR1 a lot and quickly (using something like `for i in $(seq 1 30); do kill -SIGUSR1 $(pgrep chihaya); done`), you get a fatal exit along the lines of `listen tcp 0.0.0.0:6881: bind: address already in use` from either the Prometheus server or HTTP frontend. I'm not sure what causes it yet, but most of the time it works.

### DIFF
--- a/cmd/chihaya/main.go
+++ b/cmd/chihaya/main.go
@@ -2,24 +2,125 @@ package main
 
 import (
 	"errors"
-	"net/http"
 	"os"
 	"os/signal"
 	"runtime/pprof"
 	"syscall"
 
 	log "github.com/Sirupsen/logrus"
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/cobra"
 
-	httpfrontend "github.com/chihaya/chihaya/frontend/http"
-	udpfrontend "github.com/chihaya/chihaya/frontend/udp"
+	"github.com/chihaya/chihaya/frontend/http"
+	"github.com/chihaya/chihaya/frontend/udp"
 	"github.com/chihaya/chihaya/middleware"
+	"github.com/chihaya/chihaya/pkg/prometheus"
+	"github.com/chihaya/chihaya/pkg/stop"
 	"github.com/chihaya/chihaya/storage"
 	"github.com/chihaya/chihaya/storage/memory"
 )
 
-func rootCmdRun(cmd *cobra.Command, args []string) error {
+// Run represents the state of a running instance of Chihaya.
+type Run struct {
+	configFilePath string
+	peerStore      storage.PeerStore
+	logic          *middleware.Logic
+	sg             *stop.Group
+}
+
+// NewRun runs an instance of Chihaya.
+func NewRun(configFilePath string) (*Run, error) {
+	r := &Run{
+		configFilePath: configFilePath,
+	}
+
+	return r, r.Start()
+}
+
+// Start begins an instance of Chihaya.
+func (r *Run) Start() error {
+	configFile, err := ParseConfigFile(r.configFilePath)
+	if err != nil {
+		return errors.New("failed to read config: " + err.Error())
+	}
+
+	chihayaCfg := configFile.Chihaya
+	preHooks, postHooks, err := chihayaCfg.CreateHooks()
+	if err != nil {
+		return errors.New("failed to validate hook config: " + err.Error())
+	}
+
+	r.sg = stop.NewGroup()
+	r.sg.Add(prometheus.NewServer(chihayaCfg.PrometheusAddr))
+
+	r.peerStore, err = memory.New(chihayaCfg.Storage)
+	if err != nil {
+		return errors.New("failed to create memory storage: " + err.Error())
+	}
+
+	r.logic = middleware.NewLogic(chihayaCfg.Config, r.peerStore, preHooks, postHooks)
+
+	if chihayaCfg.HTTPConfig.Addr != "" {
+		httpfe, err := http.NewFrontend(r.logic, chihayaCfg.HTTPConfig)
+		if err != nil {
+			return err
+		}
+		r.sg.Add(httpfe)
+	}
+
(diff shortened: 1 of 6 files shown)


cmd/chihaya/main.go
```diff
@@ -2,214 +2,192 @@ package main
...
+// Start begins an instance of Chihaya.
+func (r *Run) Start() error {
+	configFile, err := ParseConfigFile(r.configFilePath)
 	if err != nil {
-		return errors.New("failed to create memory storage: " + err.Error())
+		return errors.New("failed to read config: " + err.Error())
 	}
 
-	preHooks, postHooks, err := configFile.CreateHooks()
+	chihayaCfg := configFile.Chihaya
```
mrd0ll4r: I would name this `cfg` or something, looks nicer to my eyes I guess

cmd/chihaya/main.go
```diff
@@ -2,214 +2,192 @@ package main
...
-			case <-quit:
-				stop(udpFrontend, httpFrontend, logic, errChan, peerStore)
-			case <-shutdown:
-				stop(udpFrontend, httpFrontend, logic, errChan, peerStore)
-			}
-		}
-	}()
+	log.Debug("stopping logic")
+	if errs := r.logic.Stop(); len(errs) != 0 {
+		errDelimiter := "; "
```
mrd0ll4r: This can probably be made a `const` somewhere?

cmd/chihaya/main.go
```diff
@@ -2,214 +2,192 @@ package main
...
 		Short: "BitTorrent Tracker",
 		Long:  "A customizable, multi-protocol BitTorrent Tracker",
-		Run: func(cmd *cobra.Command, args []string) {
-			if err := rootCmdRun(cmd, args); err != nil {
-				log.Fatal(err)
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			debugLog, _ := cmd.Flags().GetBool("debug")
+			if debugLog {
+				log.SetLevel(log.DebugLevel)
+				log.Debugln("debug logging enabled")
```
mrd0ll4r: This is so useful, nice! :+1:

frontend/http/frontend.go
```diff
@@ -72,102 +71,103 @@ type Config struct {
...
 
 // Stop provides a thread-safe way to shutdown a currently running Frontend.
-func (t *Frontend) Stop() {
-	if err := t.s.Shutdown(context.Background()); err != nil {
-		log.Warn("Error shutting down HTTP frontend:", err)
-	}
+func (f *Frontend) Stop() <-chan error {
+	c := make(chan error)
+	go func() {
+		if err := f.srv.Shutdown(context.TODO()); err != nil {
```
mrd0ll4r: `context.Background()` ?

frontend/http/frontend.go
```diff
@@ -72,102 +71,103 @@ type Config struct {
...
+
+	return f, nil
 }
 
 // Stop provides a thread-safe way to shutdown a currently running Frontend.
-func (t *Frontend) Stop() {
-	if err := t.s.Shutdown(context.Background()); err != nil {
-		log.Warn("Error shutting down HTTP frontend:", err)
-	}
+func (f *Frontend) Stop() <-chan error {
```
mrd0ll4r: Does the HTTP frontend not need the already-stopped check like UDP does?

pkg/prometheus/server.go
```diff
@@ -0,0 +1,50 @@
...
+			c <- err
+		} else {
+			close(c)
+		}
+	}()
+
+	return c
+}
+
+// NewServer creates a new instance of an Prometheus server that asynchronously
```
mrd0ll4r: a instead of an

frontend/http/frontend.go
mrd0ll4r: Doesn't look like it does: https://play.golang.org/p/FcNbOYZxYF

PR #311, opened 2017-05-04
bittorrent: create sanitizer object

jzelinskie:
This consolidates the logic by which a request is valid into the bittorrent library rather than being implemented by individual frontends.

### DIFF
--- a/bittorrent/sanitize.go
+++ b/bittorrent/sanitize.go
@@ -21,12 +21,10 @@ type RequestSanitizer struct {
 // SanitizeAnnounce enforces a max and default NumWant and coerces the a peer's
 // IP address into the proper format.
 func (rs *RequestSanitizer) SanitizeAnnounce(r *AnnounceRequest) (*AnnounceRequest, error) {
-	if r.NumWant > rs.MaxNumWant {
-		r.NumWant = rs.MaxNumWant
-	}
-
-	if r.NumWant == 0 {
+	if !r.NumWantProvided {
 		r.NumWant = rs.DefaultNumWant
+	} else if r.NumWant > rs.MaxNumWant {
+		r.NumWant = rs.MaxNumWant
 	}
 
 	if ip := r.Peer.IP.To4(); ip != nil {
--- a/bittorrent/sanitize.go
+++ b/bittorrent/sanitize.go
@@ -4,7 +4,7 @@ import (
 	"errors"
 	"net"
 
-	log "github.com/Sirupsen/logrus"
+	log "github.com/sirupsen/logrus"
 )
 
 // ErrInvalidIP indicates an invalid IP for an Announce.
--- a/bittorrent/sanitize.go
+++ b/bittorrent/sanitize.go
@@ -1,0 +1,53 @@
+package bittorrent
+
+import (
+	"errors"
+	"net"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// ErrInvalidIP indicates an invalid IP for an Announce.
+var ErrInvalidIP = errors.New("invalid IP")
+
+// RequestSanitizer is used to replace unreasonable values in requests parsed
+// from a frontend into sane values.
+type RequestSanitizer struct {
+	MaxNumWant          uint32 `yaml:"max_numwant"`
+	DefaultNumWant      uint32 `yaml:"default_numwant"`
+	MaxScrapeInfoHashes uint32 `yaml:"max_scrape_infohashes"`
+}
+
+// SanitizeAnnounce enforces a max and default NumWant and coerces the a peer's
+// IP address into the proper format.
+func (rs *RequestSanitizer) SanitizeAnnounce(r *AnnounceRequest) (*AnnounceRequest, error) {
+	if r.NumWant > rs.MaxNumWant {
+		r.NumWant = rs.MaxNumWant
+	}
+
+	if r.NumWant == 0 {
+		r.NumWant = rs.DefaultNumWant
+	}
+
+	if ip := r.Peer.IP.To4(); ip != nil {
+		r.Peer.IP.IP = ip
+		r.Peer.IP.AddressFamily = IPv4
+	} else if len(r.Peer.IP.IP) == net.IPv6len { // implies r.Peer.IP.To4() == nil
+		r.Peer.IP.AddressFamily = IPv6
+	} else {
+		return r, ErrInvalidIP
+	}
+
+	log.Debugf("sanitized request: %#v", r)
+	return r, nil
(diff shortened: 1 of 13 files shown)


frontend/http/parser.go
```diff
@@ -63,7 +63,7 @@ func ParseAnnounce(r *http.Request, realIPHeader string, allowIPSpoofing bool) (
 	}
 
 	numwant, err := qp.Uint64("numwant")
-	if err != nil && err != bittorrent.ErrKeyNotFound {
+	if err != nil {
```
mrd0ll4r: `num_want` is still optional, isn't it?

bittorrent/sanitize.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// SanitizeAnnounce enforces a max and default NumWant and coerces the a peer's
+// IP address into the proper format.
+func (rs *RequestSanitizer) SanitizeAnnounce(r *AnnounceRequest) (*AnnounceRequest, error) {
+	if r.NumWant > rs.MaxNumWant {
+		r.NumWant = rs.MaxNumWant
+	}
+
+	if r.NumWant == 0 {
```
mrd0ll4r: Should we introduce a sentinel value to indicate `num_want` was missing rather than actually set to zero?

frontend/http/parser.go
```diff
@@ -7,12 +7,12 @@ import (
 	"github.com/chihaya/chihaya/bittorrent"
 )
 
-// ParseAnnounce parses an bittorrent.AnnounceRequest from an http.Request.
```
mrd0ll4r: This seems fine to me, esp. because you left it unchanged in L85.

frontend/http/parser.go
```diff
@@ -99,10 +99,10 @@ func ParseScrape(r *http.Request) (*bittorrent.ScrapeRequest, error) {
 		Params:     qp,
 	}
 
-	return request, nil
+	return rs.SanitizeScrape(request)
 }
 
-// requestedIP determines the IP address for a BitTorrent client request.
+// requestedIP determines the IP address for a bt client request.
```
mrd0ll4r: Let's stick with `BitTorrent`, although I've read `bt` before, it doesn't make this easier to understand.

jzelinskie: Yup this was a find/replace error

bittorrent/sanitize.go
jzelinskie: What do you think the best way to do this is?

Here's an interesting strawman:
```go
type UnsanitizedAnnounceRequest {
  NumWant MaybeUint32
  // ...
}

type MaybeNumWant struct {
  Nothing bool
  Just uint32
}

func (rs *RequestSanitizer) SanitizeAnnounce(r *UnsanitizedAnnounceRequest) (*AnnounceRequest, error) {
  // ...
}
```

frontend/http/parser.go
jzelinskie: this was overlooked when i rebased

frontend/http/parser.go
jzelinskie: changed back

bittorrent/sanitize.go
mrd0ll4r: Hmm, we have to do something like that. But I want to avoid allocating a whole new struct for it.

Maybe something like:

```go
type AnnounceRequest struct {
    NumWant OptionalNumWant
    ...
}

type OptionalNumWant struct {
    Value uint32
    Provided bool
}
```

Then sanitize accordingly and guarantee that `NumWant.Value` is set to a sane value?

Or, if we want to be really high performance and dirty, just
```go
const NumWantNotSet uint32(13337)
```
and check against that :)

I'm thinking - what are the security implications if someone can trick us, with a UDP announce, into believing no NumWant was specified? We're gonna use our default and they'll know our default, but who really cares?

jzelinskie:
Just an update on this, I rebased and I'm working at add more boolean fields to the Request struct to track optional fields. If we put the booleans after the `Compact` bool, then the request object won't change size since that bool was eating a whole byte for struct alignment.

mrd0ll4r:
> If we put the booleans after the Compact bool, then the request object won't change size since that bool was eating a whole byte for struct alignment.

I like your way of thinking :)

bittorrent/sanitize.go
```diff
@@ -0,0 +1,51 @@
...
+
+// RequestSanitizer is used to replace unreasonable values in requests parsed
+// from a frontend into sane values.
+type RequestSanitizer struct {
+	MaxNumWant          uint32 `yaml:"max_numwant"`
+	DefaultNumWant      uint32 `yaml:"default_numwant"`
+	MaxScrapeInfoHashes uint32 `yaml:"max_scrape_infohashes"`
+}
+
+// SanitizeAnnounce enforces a max and default NumWant and coerces the a peer's
```
mrd0ll4r: `and coerces the a peer's` :)

bittorrent/sanitize.go
```diff
@@ -0,0 +1,51 @@
+package bittorrent
+
+import (
+	"errors"
+	"net"
+
+	log "github.com/sirupsen/logrus"
```
mrd0ll4r: Replace with out own logging package

bittorrent/sanitize.go
```diff
@@ -0,0 +1,51 @@
...
+	if ip := r.Peer.IP.To4(); ip != nil {
+		r.Peer.IP.IP = ip
+		r.Peer.IP.AddressFamily = IPv4
+	} else if len(r.Peer.IP.IP) == net.IPv6len { // implies r.Peer.IP.To4() == nil
+		r.Peer.IP.AddressFamily = IPv6
+	} else {
+		return r, ErrInvalidIP
+	}
+
+	log.Debugf("sanitized request: %#v", r)
```
mrd0ll4r: Replace with our new and improved debug logging :)

cmd/chihaya/main.go
```diff
@@ -73,7 +73,7 @@ func (r *Run) Start(ps storage.PeerStore) error {
 
 	if cfg.HTTPConfig.Addr != "" {
 		log.Info("starting HTTP frontend", cfg.HTTPConfig.LogFields())
-		httpfe, err := http.NewFrontend(r.logic, cfg.HTTPConfig)
+		httpfe, err := http.NewFrontend(r.logic, &cfg.RequestSanitizer, cfg.HTTPConfig)
```
mrd0ll4r: Do we maybe need to check if `cfg.RequestSanitizer` is set or left empty and use defaults?

PR #322, opened 2017-06-02
dist/prometheus: add prometheus config and rules

### LINKED ISSUE #293: Prometheus records wrong address families for UDP/HTTP announces and UDP connects
We don't set the `AddressFamily`, so it always defaults to `IPv4` for announces.

mrd0ll4r:
We might wanna fix #293 first, which in turn depends on #311, because right now address families are all over the place.


dist/prometheus/prometheus.yaml
```diff
@@ -0,0 +1,14 @@
...
+rule_files:
+  - chihaya.rules
+
+# A scrape configuration containing exactly one endpoint to scrape:
+scrape_configs:
+  # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
+  - job_name: 'your-job'
+    scrape_interval: 5s # optionally override the global scrape_interval
+    static_configs:
+      - targets: ['1.2.3.4:6881'] # provide the address of chihaya's prometheus endpoint
```
jzelinskie: maybe we can make this `localhost` so it's a little more useable out of the box

dist/prometheus/prometheus.yaml
```diff
@@ -0,0 +1,14 @@
...
+  scrape_interval:     5s
+  evaluation_interval: 5s
+
+rule_files:
+  - chihaya.rules
+
+# A scrape configuration containing exactly one endpoint to scrape:
+scrape_configs:
+  # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
+  - job_name: 'your-job'
```
jzelinskie: name this local-chihaya or something

dist/prometheus/chihaya.rules
```diff
@@ -0,0 +1,74 @@
+# Chihaya recording rules
```
jzelinskie: We should definitely have documentation on what rules are (how they are optional) etc...

mrd0ll4r:
updated, ptal

jzelinskie:
LGTM. Can you add another commit that moves `contrib/` into `dist/` as well. I think it makes more sense to consolidate with that convention.

mrd0ll4r:
Will do, I can't merge this without you pressing that approve button though :)

PR #323, opened 2017-06-02
cmd/chihaya: log hook names on startup

### LINKED ISSUE #317: Improve logging of hooks in use
We now log what hooks are in use. However, we just print their addresses in memory, which in turn doesn't help a lot.

cedricgc:
Replace logging memory addresses with actual hook names.

Closes #317

### DIFF
--- a/cmd/chihaya/config.go
+++ b/cmd/chihaya/config.go
@@ -21,6 +21,18 @@ type hookConfig struct {
 	Config interface{} `yaml:"config"`
 }
 
+type hookConfigs []hookConfig
+
+// Names returns all hook names listed in the configuration.
+func (hookCfgs hookConfigs) Names() (hookNames []string) {
+	hookNames = make([]string, len(hookCfgs))
+	for index, hookCfg := range hookCfgs {
+		hookNames[index] = hookCfg.Name
+	}
+
+	return
+}
+
 // Config represents the configuration used for executing Chihaya.
 type Config struct {
 	middleware.Config `yaml:",inline"`
@@ -28,8 +40,8 @@ type Config struct {
 	HTTPConfig        httpfrontend.Config `yaml:"http"`
 	UDPConfig         udpfrontend.Config  `yaml:"udp"`
 	Storage           memory.Config       `yaml:"storage"`
-	PreHooks          []hookConfig        `yaml:"prehooks"`
-	PostHooks         []hookConfig        `yaml:"posthooks"`
+	PreHooks          hookConfigs         `yaml:"prehooks"`
+	PostHooks         hookConfigs         `yaml:"posthooks"`
 }
 
 // CreateHooks creates instances of Hooks for all of the PreHooks and PostHooks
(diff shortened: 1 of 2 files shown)


cedricgc:
Not sure if I have to rebase against latest master first, I think it should merge fine though.

cmd/chihaya/main.go
```diff
@@ -65,9 +65,17 @@ func (r *Run) Start(ps storage.PeerStore) error {
 	if err != nil {
 		return errors.New("failed to validate hook config: " + err.Error())
 	}
+	preHookNames := make([]string, 0)
+	for _, hookCfg := range cfg.PreHooks {
+		preHookNames = append(preHookNames, hookCfg.Name)
+	}
```
jzelinskie: This could be made a method of the `cfg.Prehook` and `cfg.Posthook` and then the log line will look like:

```go
log.WithFields(log.Fields{
  "prehooks": prehooks.Names(),
  "posthooks": posthooks.Names(),
}).Info("starting middleware")
```

cedricgc: will do

cedricgc: `prehooks` and `posthooks` are of type `[]middleware.Hook`, which does not have the name field like `hookConfig`. It will be more like:

```go
log.WithFields(log.Fields{
  "prehooks": cfg.PreHooks.Names(),
  "posthooks": cfg.PostHooks.Names(),
}).Info("starting middleware")
```
The `Names()` method would be for type `[]hookConfig`.

OK?

mrd0ll4r: looks good to me

cmd/chihaya/config.go
```diff
@@ -21,15 +21,27 @@ type hookConfig struct {
 	Config interface{} `yaml:"config"`
 }
 
+type hookConfigs []hookConfig
```
cedricgc: Feels awkward but not sure how else to have a method on the whole slice.

jzelinskie: nope, this is totally fine!

cmd/chihaya/config.go
```diff
@@ -21,15 +21,27 @@ type hookConfig struct {
...
+
+// Names returns all hook names listed in the configuration.
+func (hookCfgs hookConfigs) Names() (hookNames []string) {
+	hookNames = make([]string, 0)
```
jzelinskie:
```go
hookNames = make([]string, len(hookCfgs))
```

cedricgc: Adding that change makes the output look like this:


screen shot 2017-06-02 at 8 49 12 pm

cedricgc: Oh I think I need to assign to the allocated spaces instead of append now:

screen shot 2017-06-02 at 8 51 49 pm

cmd/chihaya/config.go
```diff
@@ -21,15 +21,27 @@ type hookConfig struct {
 	Config interface{} `yaml:"config"`
 }
 
+type hookConfigs []hookConfig
+
+// Names returns all hook names listed in the configuration.
+func (hookCfgs hookConfigs) Names() (hookNames []string) {
+	hookNames = make([]string, len(hookCfgs))
+	for index, hookCfg := range hookCfgs {
+		hookNames[index] = hookCfg.Name
```
cedricgc: Index instead of append now.

cedricgc:
hes done it

First open-source contribution!

Issue #319, opened 2017-05-31
Peer store panics on creation with old (a few days) config


mrd0ll4r:
This is because the prometheus reporting interval was added in the config.
If unspecified, it defaults to zero. This causes a panic in the initialization of the peer store like this:

```
panic: non-positive interval for NewTicker

goroutine 36 [running]:
time.NewTicker(0x0, 0x913b38)
	/usr/lib/go/src/time/tick.go:23 +0x1b4
github.com/chihaya/chihaya/storage/memory.New.func3(0xc420075e40, 0xc393e6d000, 0x0, 0xdf84758000, 0x400)
	<->/github.com/chihaya/chihaya/storage/memory/peer_store.go:128 +0x73
created by github.com/chihaya/chihaya/storage/memory.New
	<->/github.com/chihaya/chihaya/storage/memory/peer_store.go:140 +0x2fa
```

Fix is simple, just add a check and possible a default value of 1, same way as the encryption key used for UDP...

mrd0ll4r:
@cedricgc do you wannt to/have time to work on this?

cedricgc:
I should have time this weekend. I will follow up on IRC later.

cedricgc:
Yeah I can do this.

mrd0ll4r:
I can't assign you to this issue, but I'll be happy to review a PR :)

cedricgc:
reproduced error by removing value from config:

```bash
docker run --rm -v $(pwd):/tmp cedricgc/chihaya --config /tmp/example_config.yaml
time="2017-06-03T18:36:48Z" level=info msg="starting Prometheus server" addr="0.0.0.0:6880"
time="2017-06-03T18:36:48Z" level=info msg="starting storage" gcInterval=14m0s peerLifetime=16m0s shardCount=1024
time="2017-06-03T18:36:48Z" level=info msg="starting middleware" postHooks=[] preHooks=[]
time="2017-06-03T18:36:48Z" level=info msg="starting HTTP frontend" addr="0.0.0.0:6881" allowIPSpoofing=false enableRequestTiming=false readTimeout=5s realIPHeader=x-real-ip tlsCertPath= tlsKeyPath= writeTimeout=5s
time="2017-06-03T18:36:48Z" level=info msg="starting UDP frontend" addr="0.0.0.0:6881" allowIPSpoofing=false enableRequestTiming=false maxClockSkew=10s privateKey="paste a random string here that will be used to hmac connection IDs"
[...]

PR #325, opened 2017-06-03
storage/memory: set default prometheus_reporting_interval

cedricgc:
When left blank in the config, the default prometheus_reporting_interval value
defaulted to 0, causing a panic during peer store initialization. This
change sets the default value to 1 if not provided.

### DIFF
--- a/storage/memory/peer_store.go
+++ b/storage/memory/peer_store.go
@@ -125,6 +125,10 @@ func New(cfg Config) (storage.PeerStore, error) {
 	ps.wg.Add(1)
 	go func() {
 		defer ps.wg.Done()
+		if cfg.PrometheusReportingInterval <= 0 {
+			cfg.PrometheusReportingInterval = 1
+			log.Warn("storage: PrometheusReportingInterval not specified/invalid, defaulting to 1 second")
+		}
 		t := time.NewTicker(cfg.PrometheusReportingInterval)
 		for {
 			select {


storage/memory/peer_store.go
```diff
@@ -125,6 +125,9 @@ func New(cfg Config) (storage.PeerStore, error) {
 	ps.wg.Add(1)
 	go func() {
 		defer ps.wg.Done()
+		if cfg.PrometheusReportingInterval == 0 {
+			cfg.PrometheusReportingInterval = 1
+		}
```
cedricgc: Should I add a why comment?

storage/memory/peer_store.go
```diff
@@ -125,6 +125,9 @@ func New(cfg Config) (storage.PeerStore, error) {
 	ps.wg.Add(1)
 	go func() {
 		defer ps.wg.Done()
+		if cfg.PrometheusReportingInterval == 0 {
```
mrd0ll4r: Make this check for `<=0`. `time.Duration` is actually an `int64`, so someone could set this to a negative value.

storage/memory/peer_store.go
mrd0ll4r: I think a warning in the logs (something like "...not specified/invalid, setting to one second") would be best :)

mrd0ll4r:
Oh and please change the commit message to have "Fixes #319" instead of "Closes", by convention.

storage/memory/peer_store.go
cedricgc: will add log statement

cedricgc:
Right, since this is a bug.

PR #330, opened 2017-06-06
dist/helm: default debug and json logs

jzelinskie:


mrd0ll4r:
Sure about this? It's not ready for production with debug logging on, is it?

jzelinskie:
>Sure about this? It's not ready for production with debug logging on, is it?

Google runs with debug logs on in production ;)

We should make this as easy as possible for people to get started, if they have issues, we don't want to have to tell them to start fooling around with this stuff just to get debug data. Once they've got it configure for their instance, they can decide to make changes like this.

PR #331, opened 2017-06-06
TrackerLogic returns modified Contexts

### LINKED ISSUE #304: TrackerLogic should return modified Contexts
The `HandleAnnounce` and `HandleScrape` methods call PreHooks, which update the context given to them and return the updated version. These methods should at the end return the fully updated `context.Context`, so that frontends can pass it to `AfterAnnounce` and `AfterScrape`.

cedricgc:
Closes #304

### DIFF
--- a/frontend/frontend.go
+++ b/frontend/frontend.go
@@ -11,14 +11,20 @@ import (
 // after the response has been delivered to the client.
 type TrackerLogic interface {
 	// HandleAnnounce generates a response for an Announce.
-	HandleAnnounce(context.Context, *bittorrent.AnnounceRequest) (*bittorrent.AnnounceResponse, error)
+	//
+	// Returns the updated context, the generated AnnounceResponse and no error
+	// on success; nil and error on failure.
+	HandleAnnounce(context.Context, *bittorrent.AnnounceRequest) (context.Context, *bittorrent.AnnounceResponse, error)
 
 	// AfterAnnounce does something with the results of an Announce after it
 	// has been completed.
 	AfterAnnounce(context.Context, *bittorrent.AnnounceRequest, *bittorrent.AnnounceResponse)
 
 	// HandleScrape generates a response for a Scrape.
-	HandleScrape(context.Context, *bittorrent.ScrapeRequest) (*bittorrent.ScrapeResponse, error)
+	//
+	// Returns the updated context, the generated AnnounceResponse and no error
+	// on success; nil and error on failure.
+	HandleScrape(context.Context, *bittorrent.ScrapeRequest) (context.Context, *bittorrent.ScrapeResponse, error)
 
 	// AfterScrape does something with the results of a Scrape after it has been completed.
 	AfterScrape(context.Context, *bittorrent.ScrapeRequest, *bittorrent.ScrapeResponse)
--- a/frontend/udp/frontend.go
+++ b/frontend/udp/frontend.go
@@ -286,8 +286,9 @@ func (t *Frontend) handleRequest(r Request, w ResponseWriter) (actionName string
 		af = new(bittorrent.AddressFamily)
 		*af = req.IP.AddressFamily
 
+		var ctx context.Context
 		var resp *bittorrent.AnnounceResponse
-		resp, err = t.logic.HandleAnnounce(context.Background(), req)
+		ctx, resp, err = t.logic.HandleAnnounce(context.Background(), req)
 		if err != nil {
 			WriteError(w, txID, err)
 			return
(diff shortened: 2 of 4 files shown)


middleware/middleware.go
```diff
@@ -49,20 +49,20 @@ type Logic struct {
 }
 
 // HandleAnnounce generates a response for an Announce.
-func (l *Logic) HandleAnnounce(ctx context.Context, req *bittorrent.AnnounceRequest) (resp *bittorrent.AnnounceResponse, err error) {
+func (l *Logic) HandleAnnounce(ctx context.Context, req *bittorrent.AnnounceRequest) (context context.Context, resp *bittorrent.AnnounceResponse, err error) {
```
cedricgc: Compiler doesn't allow the same name for return variable

jzelinskie: maybe we could call them `prevCtx` and `nextCtx`?

frontend/frontend.go
```diff
@@ -11,14 +11,14 @@ import (
 // after the response has been delivered to the client.
 type TrackerLogic interface {
 	// HandleAnnounce generates a response for an Announce.
```
mrd0ll4r: Add a line about what this returns in which case, something like "returns the updated context, the generated AnnounceResponse and no error on success, ...".

frontend/frontend.go
```diff
@@ -11,14 +11,14 @@ import (
...
-	HandleAnnounce(context.Context, *bittorrent.AnnounceRequest) (*bittorrent.AnnounceResponse, error)
+	HandleAnnounce(context.Context, *bittorrent.AnnounceRequest) (context.Context, *bittorrent.AnnounceResponse, error)
 
 	// AfterAnnounce does something with the results of an Announce after it
 	// has been completed.
 	AfterAnnounce(context.Context, *bittorrent.AnnounceRequest, *bittorrent.AnnounceResponse)
 
 	// HandleScrape generates a response for a Scrape.
```
mrd0ll4r: Same here :)

frontend/frontend.go
cedricgc: Returns the updated context, the generated AnnounceResponse and no error on success; nil and error on failure.

frontend/frontend.go
```diff
@@ -10,15 +10,19 @@ import (
 // response from a parsed request, and (2) asynchronously observe anything
 // after the response has been delivered to the client.
 type TrackerLogic interface {
-	// HandleAnnounce generates a response for an Announce.
-	HandleAnnounce(context.Context, *bittorrent.AnnounceRequest) (*bittorrent.AnnounceResponse, error)
+	// HandleAnnounce generates a response for an Announce. Returns the updated
```
mrd0ll4r: Ah, sorry about the trouble, but: We usually have at least a newline, typically even an empty line before the "Returns ...", like so:

```go
// HandleAnnounce generates a response for an Announce.
//
// Returns the updated context, the generated AnnounceResponse and no
// error on success; nil and error on failure.
```
(I don't know if that fits within the line-length limit, only have a browser at hand.)

middleware/middleware.go
mrd0ll4r: You could even use the blank identifier `_` to name the return argument.

cedricgc:
will do when I am back home

cedricgc:
oh wait need to add the //

mrd0ll4r:
Alright, code looks good now! Only thing left is to squash this into one commit with a nice name and description :)

frontend/udp/frontend.go
```diff
@@ -286,16 +286,17 @@ func (t *Frontend) handleRequest(r Request, w ResponseWriter) (actionName string
 		af = new(bittorrent.AddressFamily)
 		*af = req.IP.AddressFamily
 
+		var ctx context.Context
```
jzelinskie: context.Background()

frontend/udp/frontend.go
```diff
@@ -319,16 +320,17 @@ func (t *Frontend) handleRequest(r Request, w ResponseWriter) (actionName string
 		af = new(bittorrent.AddressFamily)
 		*af = req.AddressFamily
 
+		var ctx context.Context
```
jzelinskie: context.Background()

frontend/udp/frontend.go
mrd0ll4r: I don't think so - this is the same as with `resp` on the next line. They're both gonna be set assigned to in line 291.

cedricgc: I did it to match resp; I can change the assignment to set the value on declaration.

jzelinskie: Yeah, this was a nitpick, but i don't think it really matters at all.

PR #335, opened 2017-06-07
middleware/jwt: encode infohashes as hex

jzelinskie:
This avoids some corner-cases that can be experienced when trying to
url-escape raw bytes.


mrd0ll4r:
This could use some documentation at some point, to describe what the JWT has to look like. And maybe issues like key rotation etc.

PR #336, opened 2017-06-10
pkg/prand: replace with pkg/xorshift

### LINKED ISSUE #312: varinterval middleware doesn't require true randomness
This middleware is expensive in terms of performance and doesn't need true randomness in order to spread out load.

### LINKED ISSUE #328: Replace prand package with xorshift
* [Potential library](https://github.com/lazybeaver/xorshift)
* [Wikipedia article](https://en.wikipedia.org/wiki/Xorshift)

We could also use a port of the [RNG used in Mario Kart 64](https://gist.github.com/jzelinskie/af9a119e697c3c8a0dae3e388d60310b) lol

mrd0ll4r:
I chose to have 1024 instances and lock them individually, because I think that's the fastest and cleanest solution.
We could use a `sync.Pool` instead of locking them, but that introduces a lot of conversions from/to interfaces.
We could also allocate a new `XORShift128Plus` from two bytes of the infohash for every request, but that produces a lot of garbage.

Fixes #328 
Fixes #312


jzelinskie:
The commit/PR title is a little incorrect, but the change looks fine

mrd0ll4r:
Oh, yeah. Fixed :)

jzelinskie:
>that produces a lot of garbage

Does it? Can we just allocate that on the stack?

mrd0ll4r:
We could, yeah, but I don't think we'd even need xorshift then. The random number produced is just `state[0] + state[1]`, after which the states are advanced. But we might as well use `infohash[0] + infohash[1]`, without any shifting or state...

Now that I think about it, maybe we should? We're always gonna have an infohash to derive things from and potentially a PeerID, if the Client provides it. If we're really cool, we could even take something from the `n`-th peer returned, because that should be changing over time even for the same Peer/Infohash combination.

Thoughts?

jzelinskie:
Yeah, I don't think this necessarily needs to be random, it could be derived from the request data.

mrd0ll4r:
I will make this happen.

PR #337, opened 2017-06-17
bittorrent: make invalid query escape errors static

### LINKED ISSUE #326: Prometheus still records invalid URL escapes
... which inflates the metrics.

I'm not sure where to fix this. We could have a static error in the `bittorrent` package that is used when URL query parsing fails. But that kind of hurts debugging.

mrd0ll4r:
Fixes #326


mrd0ll4r:
#338 is better, closing this

PR #338, opened 2017-06-17
Fix parsing

### LINKED ISSUE #326: Prometheus still records invalid URL escapes

mrd0ll4r:
Fixes #337 
Fixes #326 

Makes URL parsing about 15% faster on my machine:
```
BenchmarkParseQuery       	 2000000	      2016 ns/op	     387 B/op	       3 allocs/op
BenchmarkParseQuery       	 2000000	      2013 ns/op	     387 B/op	       3 allocs/op
BenchmarkParseQuerySimple 	 3000000	      1707 ns/op	     387 B/op	       3 allocs/op
BenchmarkParseQuerySimple 	 3000000	      1703 ns/op	     387 B/op	       3 allocs/op
BenchmarkURLParseQuery    	 2000000	      1852 ns/op	     463 B/op	       7 allocs/op
BenchmarkURLParseQuery    	 2000000	      1847 ns/op	     463 B/op	       7 allocs/op
```

`ParseQuery` is what we had before, `ParseQuerySimple` is this (a simplification of `url.parseQuery`), `URLParseQuery` is `url.ParseQuery`.

### DIFF
--- a/bittorrent/params.go
+++ b/bittorrent/params.go
@@ -88,69 +88,45 @@ func ParseURLData(urlData string) (*QueryParams, error) {
 
 // parseQuery parses a URL query into QueryParams.
 // The query is expected to exclude the delimiting '?'.
-func parseQuery(rawQuery string) (*QueryParams, error) {
-	var (
-		keyStart, keyEnd int
-		valStart, valEnd int
+func parseQuery(query string) (q *QueryParams, err error) {
+	// This is basically url.parseQuery, but with a map[string]string
+	// instead of map[string][]string for the values.
+	q = &QueryParams{
+		query:      query,
+		infoHashes: nil,
+		params:     make(map[string]string),
+	}
 
-		onKey = true
-
-		q = &QueryParams{
-			query:      rawQuery,
-			infoHashes: nil,
-			params:     make(map[string]string),
-		}
-	)
-
-	for i, length := 0, len(rawQuery); i < length; i++ {
-		separator := rawQuery[i] == '&' || rawQuery[i] == ';'
-		last := i == length-1
-
-		if separator || last {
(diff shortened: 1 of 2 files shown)


bittorrent/params_test.go
```diff
@@ -27,6 +27,12 @@ var (
 	InvalidQueries = []string{
 		"/announce?" + "info_hash=%0%a",
 	}
+
+	// See https://github.com/chihaya/chihaya/issues/334.
+	panicQueries = []string{
```
jzelinskie: This seems to imply that these queries _should_ panic

bittorrent/params.go
```diff
@@ -88,69 +94,55 @@ func ParseURLData(urlData string) (*QueryParams, error) {
...
+		if i := strings.Index(key, "="); i >= 0 {
+			key, value = key[:i], key[i+1:]
+		}
+		key, err = url.QueryUnescape(key)
+		if err != nil {
+			// QueryUnescape returns an error like "invalid escape: '%x'".
+			// But frontends record these errors to prometheus, which generates
+			// a lot of time series.
+			// We log it here for debugging instead.
+			log.Debugln(err)
```
jzelinskie: can we make this fit the newer log format?

```go
log.WithFields(log.Fields{"error": err}).Debug("failed to unescape query param")
```

mrd0ll4r: But that's gonna degrade performance even with debug logging disabled.

jzelinskie: ok, i guess we'll save that for a logging package PR

PR #339, opened 2017-06-17
pkg/xorshift: rebuild to use stack only

mrd0ll4r:
Follow-up to #336


middleware/varinterval/varinterval.go
```diff
@@ -62,23 +61,27 @@ func New(cfg Config) (middleware.Hook, error) {
...
-		h.pr[i] = xorshift.NewLockedXORShift128Plus(rand.Uint64(), rand.Uint64())
 	}
 	return h, nil
 }
 
-func (h *hook) getXORShiftByInfohash(ih *bittorrent.InfoHash) *xorshift.LockedXORShift128Plus {
-	return h.pr[(int(ih[1])|int(ih[0])<<8)%len(h.pr)]
+// getStateFromRequest generates 2*64 bits of pseudo random state from an
+// AnnounceRequest.
+func (h *hook) getStateFromRequest(req *bittorrent.AnnounceRequest) (uint64, uint64) {
```
jzelinskie: this doesn't need to be a method, right? i don't see it using the hook anywhere

jzelinskie:
I think xorshift stuff and the DeriveEntropy func should in the same package `middleware/pkg/random`

mrd0ll4r:
Makes sense, ptal

PR #343, opened 2017-06-25
storage: update config defaults

### LINKED ISSUE #341: Set valid PeerLifetime in default storage configs
If, for example due to a malformed config file, the default config for a storage is used, `PeerLifetime` will be set to zero. This will collect all peers on every GC.

mrd0ll4r:
Some cleanup, probably more sane defaults.

Fixes #341

### DIFF
--- a/storage/memory/peer_store.go
+++ b/storage/memory/peer_store.go
@@ -6,7 +6,6 @@ import (
 	"encoding/binary"
 	"net"
 	"runtime"
-	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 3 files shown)


storage/memory/peer_store.go
```diff
@@ -70,19 +69,17 @@ func (cfg Config) LogFields() log.Fields {
...
-			"provided": strconv.Itoa(cfg.ShardCount),
-			"default":  strconv.Itoa(validcfg.ShardCount),
+			"provided": cfg.ShardCount,
+			"default":  validcfg.ShardCount,
 		}).Warnln("falling back to default configuration")
 	}
 
 	if cfg.GarbageCollectionInterval <= 0 {
-		validcfg.GarbageCollectionInterval = time.Minute * 14
+		validcfg.GarbageCollectionInterval = time.Minute * 3
```
jzelinskie: We should probably pull these defaults out into constants

mrd0ll4r:
updated, ptal

PR #353, opened 2017-08-26
add retracker.local compatibility

askovpen:


mrd0ll4r:
Please add an explanation about this.

jzelinskie:
I actually don't mind including this patch if it just looked like this:

```go
	router.GET("/announce", f.announceRoute)
  	router.GET("/scrape", f.scrapeRoute)
 +
 +      // These routes are for compatibility with legacy PHP trackers.
 +	router.GET("/announce.php", f.announceRoute)
 +	router.GET("/scrape.php", f.scrapeRoute)
```

jonasbits:
is this in regard to this BEP http://bittorrent.org/beps/bep_0022.html

mrd0ll4r:
Yes and no.

The idea is the same: Prefer local peers for up/downloading. This can be achieved by having a 'local' tracker run inside the local network that  returns peers from that network, or by having trackers return peers that share the same IP prefix with you (what facebook does). The advantage of the whole thing is, of course, to have higher up/download speed and less peering traffic between ASes.

The BEP is the 'nice' way of doing it: Use DNS to find and resolve a local tracker, starting from very local, expanding to ever-less local.
The `retracker.local` approach is a little more dirty: Just add `retracker.local` to all your torrents and perform standard discovery/lookup of whoever is named `retracker.local`, then use that machine as a local tracker. See the two attached wiki articles from rutracker for an explanation.

For the BEP we don't have to change anything. For the `retracker.local` business we _should_ not have to change anything either, the articles linked below recommend doing URL rewrites with nginx and running xbt or opentracker, so that the actual announce URL in the torrent files ends with `/announce`, instead of `/announce.php`. (Our announce URL is `/announce` already, hence we don't have to change anything.)

https://translate.google.de/translate?hl=en&sl=ru&u=http://rutracker.wiki/Оптимизация_обмена_битторрент_траффиком_в_локальных_сетях
https://translate.google.de/translate?hl=en&sl=ru&u=http://rutracker.wiki/Retracker.local

My verdict: Add a config key to enable legacy PHP URLs. And add a comment to it explaining why this is bad. :)

mrd0ll4r:
Closing, fixed by #373

PR #362, opened 2017-09-30
http: add default ReadTimeout, WriteTimeout

### LINKED ISSUE #359: Default Read/Write Timeouts
We need default read/write timeouts for HTTP, in case the user forgets to set them.

Not a problem for UDP, I guess.

mrd0ll4r:
Fixes #359

### DIFF
--- a/frontend/http/frontend.go
+++ b/frontend/http/frontend.go
@@ -86,6 +86,40 @@ func (cfg Config) LogFields() log.Fields {
 	}
 }
 
+// Default config constants.
+const (
+	defaultReadTimeout  = 2 * time.Second
+	defaultWriteTimeout = 2 * time.Second
+)
+
+// Validate sanity checks values set in a config and returns a new config with
+// default values replacing anything that is invalid.
+//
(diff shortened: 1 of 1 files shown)


jzelinskie:
I'm pretty sure we removed these because they were a source of a lot of calls to `gettimeofday`.
I think this change is going to require profiling.

mrd0ll4r:
We removed them for UDP in #300. We can't remove them in HTTP because they're built into the server and, as testing showed, actually pretty useful. (Goroutines explode into the 100k when these are unset) See also [this cloudflare blog](https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/) about setting timeouts.

What I want to do ideally is set the WriteTimeout in each handler so we could use the cached clock to avoid calls to `time.Now()`, an issue proposing something like that is [in the works](https://github.com/golang/go/issues/16100#issuecomment-328326943), but that might take until at least go1.10 to be finished.

So the point, really, is to set them to something `!= 0` to get rid of broken clients.

PR #363, opened 2017-10-08
Request Sanitizer via library

jzelinskie:
This PR deprecates #311 

These changes expose the request sanitizer as a library rather than a middleware.
It introduces new fields to determine the original presence of values in the AnnounceRequest struct. However, these new fields do not take up anymore memory by reusing memory that would otherwise have been padded out due to struct alignment.

Each server has its own sanitizer because I do foresee users wanted tighter restrictions on a per-protocol basis.


frontend/http/parser.go
```diff
@@ -7,28 +7,43 @@ import (
...
-// If allowIPSpoofing is true, IPs provided via params will be used.
-// If realIPHeader is not empty string, the first value of the HTTP Header with
+// If AllowIPSpoofing is true, IPs provided via BitTorrent params will be used.
+// If RealIPHeader is not empty string, the value of the first HTTP Header with
 // that name will be used.
-func ParseAnnounce(r *http.Request, realIPHeader string, allowIPSpoofing bool) (*bittorrent.AnnounceRequest, error) {
+type ParseOptions struct {
+	AllowIPSpoofing             bool   `yaml:"allowIPSpoofing"`
+	RealIPHeader                string `yaml:"realIPHeader"`
+	bittorrent.RequestSanitizer `yaml:",inline"`
```
mrd0ll4r: Uh, this looks a bit weird, like the actual sanitizer is embedded in this options struct, which is eventually gonna be a file.

frontend/http/parser.go
```diff
@@ -47,43 +63,54 @@ func ParseAnnounce(r *http.Request, realIPHeader string, allowIPSpoofing bool) (
...
 		return nil, bittorrent.ClientError("failed to parse parameter: uploaded")
 	}
 
+	// Determine the number of peers the client wants in the response.
 	numwant, err := qp.Uint64("numwant")
 	if err != nil && err != bittorrent.ErrKeyNotFound {
 		return nil, bittorrent.ClientError("failed to parse parameter: numwant")
 	}
 	request.NumWant = uint32(numwant)
+	request.NumWantProvided = err == nil
```
mrd0ll4r: Maybe add a comment why `err == nil` means NumWant  was provided. I'm  sure someone is gonna scratch his head about this in the future.

bittorrent/request_sanitizer.go
```diff
@@ -0,0 +1,61 @@
...
+// from a frontend into sane values.
+type RequestSanitizer struct {
+	MaxNumWant          uint32 `yaml:"max_numwant"`
+	DefaultNumWant      uint32 `yaml:"default_numwant"`
+	MaxScrapeInfoHashes uint32 `yaml:"max_scrape_infohashes"`
+}
+
+// SanitizeAnnounce enforces a max and default NumWant and coerces the peer's
+// IP address into the proper format.
+func (rs *RequestSanitizer) SanitizeAnnounce(r *AnnounceRequest) error {
```
mrd0ll4r: An idea about this:
The method doesn't need to be defined on a pointer to the struct, it could be defined on the struct itself. (Because it doesn't change it and the struct is pretty small.) And if we do that, we could also just pass it as a parameter. Then we could rename `RequestSanitizer` to something like `RequestSanitizerOptions`, which would also make the config structs further down nicer.

Thoughts?

frontend/http/parser.go
jzelinskie: sounds good.

frontend/http/parser.go
jzelinskie: The Sanitizer is embedded in the ParseOptions struct. I'm not sure what you're saying when you say "which is eventually going to be a file". Can you rephrase your thought?

mrd0ll4r: Because it's serialized and deserialized to a file, and it looks like we have this active-ey Sanitizer thing in it. Doesn't look like options, if that makes sense?

frontend/http/parser.go
```diff
@@ -47,43 +65,55 @@ func ParseAnnounce(r *http.Request, realIPHeader string, allowIPSpoofing bool) (
 	}
 	request.Peer.ID = bittorrent.PeerIDFromString(peerID)
 
+	// Determine the number of remaining pieces for the client.
```
mrd0ll4r: All of these (left, downloaded, uploaded) are numbers of bytes, not pieces, iirc.

Sorry, I should've noticed with the last review.

PR #368, opened 2017-12-02
udp: set address family for connect metrics

### LINKED ISSUE #293: Prometheus records wrong address families for UDP/HTTP announces and UDP connects

mrd0ll4r:
Fixes #293

### DIFF
--- a/frontend/udp/frontend.go
+++ b/frontend/udp/frontend.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
+	"fmt"
 	"math/rand"
 	"net"
 	"sync"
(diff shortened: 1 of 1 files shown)


frontend/udp/frontend.go
```diff
@@ -273,6 +273,18 @@ func (t *Frontend) handleRequest(r Request, w ResponseWriter) (actionName string
 			return
 		}
 
+		af = new(bittorrent.AddressFamily)
+		if r.IP.To4() != nil {
+			*af = bittorrent.IPv4
+		} else if len(r.IP) == net.IPv6len { // implies r.IP.To4() == nil
+			*af = bittorrent.IPv6
+		} else {
+			// Should never happen - we got the IP straight from the UDP packet.
```
jzelinskie: If this should never happen, this should be a `panic`.

mrd0ll4r:
makes sense, ptal

frontend/udp/frontend.go
```diff
@@ -273,6 +274,16 @@ func (t *Frontend) handleRequest(r Request, w ResponseWriter) (actionName string
...
+			panic(fmt.Sprintf("udp: invalid IP: neither v4 nor v6, IP: %+v", r.IP))
```
jzelinskie: nit, i prefer %#v

mrd0ll4r:
ptal

PR #372, opened 2017-12-05
cmd/chihaya: make things work on windows

### LINKED ISSUE #302: Build on Windows fails
... because Windows does not have `SIGUSR` signals. It has `SIGHUP`, we could abuse that, but I don't like it. Maybe use build tags to use `SIGHUP` on Windows and `SIGUSR1` on any sane system?

mrd0ll4r:
Fixes #302

Also makes it so that logging on windows has coloring disabled by default.

Before:
```
?[36mINFO?[0m[0000] starting Prometheus server                    ?[36maddr?[0m="0.0.0.0:6880"
?[36mINFO?[0m[0000] starting storage                              ?[36mname?[0m=memory
?[36mINFO?[0m[0000] started storage                               ?[36mgcInterval?[0m=3m0s ?[36mname?[0m=memory ?[36mpeerLifetime?[0m=16m0s ?[36mpromReportInterval?[0m=1s ?[36mshardCount?[0m=1024
?[36mINFO?[0m[0000] starting middleware                           ?[36mpostHooks?[0m="[]" ?[36mpreHooks?[0m="[]"
?[36mINFO?[0m[0000] starting HTTP frontend                        ?[36maddr?[0m="0.0.0.0:6881" ?[36mallowIPSpoofing?[0m=true ?[36mdefaultNumWant?[0m=50 ?[36menableRequestTiming?[0m=true ?[36mmaxNumWant?[0m=100 ?[36mmaxScrapeInfoHashes?[0m=50 ?[36mreadTimeout?[0m
=5s ?[36mrealIPHeader?[0m= ?[36mtlsCertPath?[0m= ?[36mtlsKeyPath?[0m= ?[36mwriteTimeout?[0m=5s
?[36mINFO?[0m[0000] starting UDP frontend                         ?[36maddr?[0m="0.0.0.0:6881" ?[36mallowIPSpoofing?[0m=true ?[36mdefaultNumWant?[0m=50 ?[36menableRequestTiming?[0m=true ?[36mmaxClockSkew?[0m=10s ?[36mmaxNumWant?[0m=100 ?[36mmaxScrapeInfoHashes?[
0m=50 ?[36mprivateKey?[0m="paste a random string here that will be used to hmac connection IDs"
?[36mINFO?[0m[0001] shutting down; received SIGINT/SIGTERM
```

After:
```
time="2017-12-05T10:08:03+01:00" level=info msg="starting Prometheus server" addr="0.0.0.0:6880"
time="2017-12-05T10:08:03+01:00" level=info msg="starting storage" name=memory
time="2017-12-05T10:08:03+01:00" level=info msg="started storage" gcInterval=3m0s name=memory peerLifetime=16m0s promReportInterval=1s shardCount=1024
time="2017-12-05T10:08:03+01:00" level=info msg="starting middleware" postHooks="[]" preHooks="[]"
time="2017-12-05T10:08:03+01:00" level=info msg="starting HTTP frontend" addr="0.0.0.0:6881" allowIPSpoofing=true defaultNumWant=50 enableRequestTiming=true maxNumWant=100 maxScrapeInfoHashes=50 readTimeout=5s realIPHeader= tlsCertPath= tlsKeyPath= writeTimeout=
5s
time="2017-12-05T10:08:03+01:00" level=info msg="starting UDP frontend" addr="0.0.0.0:6881" allowIPSpoofing=true defaultNumWant=50 enableRequestTiming=true maxClockSkew=10s maxNumWant=100 maxScrapeInfoHashes=50 privateKey="paste a random string here that will be
 used to hmac connection IDs"
time="2017-12-05T10:08:07+01:00" level=info msg="shutting down; received SIGINT/SIGTERM"
```

I have no clue why the entire format changed - I just disabled colors...

### DIFF
--- a/cmd/chihaya/main.go
+++ b/cmd/chihaya/main.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"os"
 	"os/signal"
+	"runtime"
 	"runtime/pprof"
 	"strings"
 	"syscall"
@@ -141,8 +142,7 @@ func RunCmdFunc(cmd *cobra.Command, args []string) error {
 	quit := make(chan os.Signal)
 	signal.Notify(quit, syscall.SIGINT, syscall.SIGTERM)
 
-	reload := make(chan os.Signal)
-	signal.Notify(reload, syscall.SIGUSR1)
+	reload := makeReloadChan()
 
 	for {
 		select {
@@ -173,6 +173,14 @@ func main() {
 		Short: "BitTorrent Tracker",
 		Long:  "A customizable, multi-protocol BitTorrent Tracker",
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			noColors, err := cmd.Flags().GetBool("nocolors")
+			if err != nil {
+				return err
+			}
+			if noColors {
+				log.SetFormatter(&logrus.TextFormatter{DisableColors: true})
+			}
+
 			jsonLog, err := cmd.Flags().GetBool("json")
 			if err != nil {
 				return err
@@ -216,6 +224,11 @@ func main() {
 	rootCmd.Flags().String("cpuprofile", "", "location to save a CPU profile")
 	rootCmd.Flags().Bool("debug", false, "enable debug logging")
 	rootCmd.Flags().Bool("json", false, "enable json logging")
+	if runtime.GOOS == "windows" {
+		rootCmd.Flags().Bool("nocolors", true, "disable log coloring")
+	} else {
+		rootCmd.Flags().Bool("nocolors", false, "disable log coloring")
+	}
 
 	if err := rootCmd.Execute(); err != nil {
 		log.Fatal("failed when executing root cobra command: " + err.Error())
--- a/cmd/chihaya/signal_unix.go
+++ b/cmd/chihaya/signal_unix.go
@@ -1,0 +1,15 @@
+// +build darwin freebsd linux netbsd openbsd dragonfly solaris
+
+package main
+
+import (
+	"os"
+	"os/signal"
+	"syscall"
+)
+
+func makeReloadChan() <-chan os.Signal {
+	reload := make(chan os.Signal)
+	signal.Notify(reload, syscall.SIGUSR1)
+	return reload
+}
--- a/cmd/chihaya/signal_windows.go
+++ b/cmd/chihaya/signal_windows.go
@@ -1,0 +1,15 @@
+// +build windows
+
+package main
+
+import (
+	"os"
+	"os/signal"
+	"syscall"
+)
+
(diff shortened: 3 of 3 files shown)


cmd/chihaya/main.go
```diff
@@ -173,6 +173,14 @@ func main() {
 		Short: "BitTorrent Tracker",
 		Long:  "A customizable, multi-protocol BitTorrent Tracker",
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			noColors, err := cmd.Flags().GetBool("nocolors")
+			if err != nil {
+				return err
+			}
+			if noColors {
+				log.SetFormatter(&logrus.TextFormatter{DisableColors: true})
```
jzelinskie: When you replace the TextFormatter with this new instance, that's what's clearing the formatting.

mrd0ll4r: But it just uses an empty `TextFormatter` by default, see [here](https://github.com/sirupsen/logrus/blob/master/logger.go#L69-L76). I give it the same one, disable colors, and the whole format changes. But as far as I can see from the source, that's intended behaviour...

jzelinskie: Looks like this is just hardcoded: [Colored format](https://github.com/sirupsen/logrus/blob/d682213848ed68c0a260ca37d6dd5ace8423f5ba/text_formatter.go#L134), [Uncolored format](https://github.com/sirupsen/logrus/blob/d682213848ed68c0a260ca37d6dd5ace8423f5ba/text_formatter.go#L103)

mrd0ll4r: I'm lost, what should be fixed upstream?
I'm okay with the uncolored format it outputs, was just surprised it did that.

jzelinskie: Two fixes for logrus upstream could be proposed:

1. logrus internally has a function called [checkIfTerminal](https://github.com/sirupsen/logrus/blob/d682213848ed68c0a260ca37d6dd5ace8423f5ba/text_formatter.go#L64), it should probably have another function for detecting Windows and not printing terminal escape sequences when doing so.
2. the non-colored output should be the same as the colored output. If you look at the implementation, they literally just use two different format strings.

mrd0ll4r: Funny enough, they claim this to be [fixed](https://github.com/sirupsen/logrus/issues/172). I just checked, `TERM` is not set for me.

So, uh, is that blocking the PR? I can remove the messing with the logging for now, if that helps.

jzelinskie: Are you sure we just don't have an old version of logrus vendored?
I'm not a huge fan of the flag for colors, I'd rather leave that to be detected by logrus.

PR #374, opened 2017-12-12
Improve Dockerfile

ilyaglow:
Hello there!

This PR drastically reduces image size using multistaged docker build and improves security: container starts as a non-root user and there is nothing except binary and yaml config inside a scratch container. 

Aside from this I changed deprecated keyword `MAINTAINER` to `LABEL` and changed `ADD` to `COPY` as suggested by [docker best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy).

Hope you'll find it useful.

Cheers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
-FROM golang:alpine
-MAINTAINER Jimmy Zelinskie <jimmyzelinskie@gmail.com>
+FROM golang:alpine AS build-env
+LABEL maintainer "Jimmy Zelinskie <jimmyzelinskie@gmail.com>"
 
 # Install OS-level dependencies.
 RUN apk update && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -8,16 +8,21 @@ RUN apk update && \
 
 # Copy our source code into the container.
 WORKDIR /go/src/github.com/chihaya/chihaya
-ADD . /go/src/github.com/chihaya/chihaya
+COPY . /go/src/github.com/chihaya/chihaya
 
 # Install our golang dependencies and compile our binary.
 RUN glide install
-RUN go install github.com/chihaya/chihaya/cmd/chihaya
+RUN CGO_ENABLED=0 GOOS=linux go install github.com/chihaya/chihaya/cmd/chihaya
```
mrd0ll4r: I've not used docker before, ever, but doesn't it work on more than just linux?

mrd0ll4r:
Thanks for the PR!
@jzelinskie is the docker person :)

Dockerfile
ilyaglow: Nice catch! Removed, thanks.

Dockerfile
```diff
@@ -1,5 +1,5 @@
-FROM golang:alpine
-MAINTAINER Jimmy Zelinskie <jimmyzelinskie@gmail.com>
+FROM golang:alpine AS build-env
+LABEL maintainer "Jimmy Zelinskie <jimmyzelinskie@gmail.com>"
```
jzelinskie: Any reason why you prefer this?

Dockerfile
```diff
@@ -8,16 +8,21 @@ RUN apk update && \
...
 # Install our golang dependencies and compile our binary.
 RUN glide install
-RUN go install github.com/chihaya/chihaya/cmd/chihaya
+RUN CGO_ENABLED=0 go install github.com/chihaya/chihaya/cmd/chihaya
+RUN adduser -D chihaya
 
-# Delete the compiler from the container.
-# This makes the container much smaller when using Quay's squashing feature.
-RUN rm -r /usr/local/go
+FROM scratch
```
jzelinskie: I've been burned by using scratch with Go binaries in the past, if you aren't extremely careful about how you compile. For example, by default Go links to libc and uses it for networking (iirc DNS resolution), so putting a binary in scratch without compiling with the `netgo` tag passed to the linker can cause subtle problems.

We can just use `alpine:latest` for the runtime container to avoid some of this complexity and so that we're running the same environment in which it was built, just without the filesystem overhead of the development toolchain.

ilyaglow: Scratch image works pretty well for me, but I don't mind to do alpine based Dockerfile in a day or two.

jzelinskie: Yeah, I suppose Chihaya doesn't use any DNS resolution, but there's a chance in the future some middleware could use it.

ilyaglow: Forgot to thank you for sharing this, didn’t know about that caveat of scratch images :) So if I just change “FROM scratch” to “FROM alpine” we will be fine?

PR #377, opened 2017-12-23
vendor: migrate from glide to dep

jzelinskie:


.travis.yml
```diff
@@ -1,16 +1,39 @@
...
+    - go get -t ./...
+    - go get -u github.com/golang/lint/...
+    - go get -u golang.org/x/tools/cmd/...
+    script:
+    - go test -v $(go list ./... | grep -v /vendor/)
+    - go vet $(go list ./... | grep -v /vendor/)
+    - diff <(goimports -d $(find . -type f -name '*.go' -not -path "./vendor/*")) <(printf "")
+    - (for d in $(go list ./... | grep -v /vendor/); do diff <(golint $d) <(printf "") || exit 1;  done)
+  allow_failures:
+  # Using HEAD of dependencies
```
mrd0ll4r: Why do we have this once with `allow_failures` and once without?

.travis.yml
```diff
@@ -1,16 +1,39 @@
...
+    - go get -u golang.org/x/tools/cmd/...
+    script:
+    - dep ensure
+    - go test -v $(go list ./... | grep -v /vendor/)
+    - go vet $(go list ./... | grep -v /vendor/)
+    - diff <(goimports -d $(find . -type f -name '*.go' -not -path "./vendor/*")) <(printf "")
+    - (for d in $(go list ./... | grep -v /vendor/); do diff <(golint $d) <(printf "") || exit 1;  done)
+  # Using HEAD of dependencies
+  - install:
+    - go get -t ./...
```
mrd0ll4r: I don't know how the travis `matrix` stuff works, do we get a completely clean environment each time? Do we maybe need to delete the `vendor/` directory from the run above? Or do we need a `-t -u` here to update all the dependencies?

Gopkg.toml
```diff
@@ -0,0 +1,58 @@
...
+  name = "github.com/spf13/cobra"
+  version = "0.0.1"
+
+[[constraint]]
+  name = "github.com/stretchr/testify"
+  version = "1.1.4"
+
+[[constraint]]
+  branch = "v2"
+  name = "gopkg.in/yaml.v2"
```
mrd0ll4r: Oh that's interesting, it specifies the branch. So we could actually use `github.com/go-yaml/yaml` as the dependency? What's your take on that?

mrd0ll4r:
I have too little knowledge of dep to make a decision rn, but I added a bunch of questions :)

.travis.yml
jzelinskie: This is just how failure matrix works on TravisCI.

.travis.yml
jzelinskie: The array items in the `include` object overwrite anything defined top-level, but give you a totally separate build for it. You have to copy them exactly into the failure matrix object, in order to have them built, but ignored for overall success.

Gopkg.toml
jzelinskie: We could! That sounds like a great follow up change.

PR #378, opened 2017-12-24
travis: update go version

dvrkps:


dvrkps:
@mrd0ll4r yap 1.x is also option. would you like that i change 1.9.x to 1.x or leave this as is ?

mrd0ll4r:
Fine with me either way - @jzelinskie opinion?

jzelinskie:
`go1.8` for example is also short for `go1.8.x`.
If we just remove this whole block and leave `language: go` it should do the same.

dvrkps:
@jzelinskie @mrd0ll4r done.

mrd0ll4r:
@jzelinskie have you tried that somewhere before? The minimal example on travis says
```
  language: go
  go:
    - 1.6
```

Even though travis builds the PR, it uses the `travis.yml` from the repository, not the PR, so I wouldn't really trust that.

dvrkps:
@jzelinskie @mrd0ll4r without go key travis use go 1.9 instead 1.9.2 [build without go key](https://travis-ci.org/chihaya/chihaya/jobs/321141354)
may i add 1.9.x ?

mrd0ll4r:
Oh my bad, it does use the updated `travis.yml` then. @dvrkps yes, make it `1.9.x` please. Sorry for the confusion.

dvrkps:
@mrd0ll4r done.

PR #381, opened 2017-12-29
*: add subtests for all table driven tests

jzelinskie:
Because we use testify, this is less useful than normal, but this is
still best practice for table-driven tests.

### DIFF
--- a/bittorrent/event_test.go
+++ b/bittorrent/event_test.go
@@ -23,10 +23,17 @@ func TestNew(t *testing.T) {
 	}
 
 	for _, tt := range table {
-		t.Run(fmt.Sprintf("%#v expecting %s", tt.data, tt.expectedErr), func(t *testing.T) {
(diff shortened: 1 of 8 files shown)


bittorrent/event_test.go
```diff
@@ -22,8 +23,10 @@ func TestNew(t *testing.T) {
 	}
 
 	for _, tt := range table {
-		got, err := NewEvent(tt.data)
-		require.Equal(t, err, tt.expectedErr, "errors should equal the expected value")
-		require.Equal(t, got, tt.expected, "events should equal the expected value")
+		t.Run(fmt.Sprintf("%#v expecting %s", tt.data, tt.expectedErr), func(t *testing.T) {
```
mrd0ll4r: The expected `err` is nil most of the time, looks pretty weird in the travis log, see [here](https://travis-ci.org/chihaya/chihaya/jobs/323034858#L555)

PR #382, opened 2018-01-03
travis: add chihaya e2e testing

### LINKED ISSUE #151: Integrate 'Poke' tool into continuous integration
At the moment it's not tested at all

mrd0ll4r:
Example travis log [looks a little cluttered](https://travis-ci.org/chihaya/chihaya/jobs/324691145) but seems to work.

Fixes #151

### DIFF
--- a/cmd/chihaya-e2e/main.go
+++ b/cmd/chihaya-e2e/main.go
@@ -1,0 +1,127 @@
+package main
+
+import (
+	"crypto/rand"
+	"flag"
+	"fmt"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/anacrolix/torrent/tracker"
+	"github.com/pkg/errors"
+
+	"github.com/chihaya/chihaya/bittorrent"
+)
+
+func init() {
+	flag.StringVar(&httpTrackerURL, "http", "http://localhost:6881/announce", "the address of the HTTP tracker")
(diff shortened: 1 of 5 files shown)


jzelinskie:
I wonder if this is something we should just vendor into the repository and name it like `cmd/chihaya-e2e`.

mrd0ll4r:
I thought about that too, but then we have a direct dependency on `github.com/astaxie/torrent`, which to me looks a little weird, given that we have the `bittorrent` package which has type definitions for all the bittorrent-related things too. We just don't have a mechanism to perform announces.

mrd0ll4r:
Alternative idea: What if we put it in github.com/chihaya/ekop ?

jzelinskie:
tbh, i'm kinda freaked out by the fact that the bencode repo has to live there forever to make the builds for 1.0 reproducible...

mrd0ll4r:
Oh, I never considered that. Alright, I'll move it to `cmd/chihaya-e2e` then

cmd/chihaya-e2e/main.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"crypto/rand"
+	"flag"
+	"fmt"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/anacrolix/torrent/tracker"
+	"github.com/pkg/errors"
```
jzelinskie: you need to add these to dep

mrd0ll4r:
alright, ptal

PR #389, opened 2018-02-09
Add Support for Redis Storage Backend

duyanghao:
This PR adds support for redis storage backend and `Chihaya` can achieve `high availability` by using [redis](https://redis.io/) storage backend.

test:
```
# go test -bench $(go list ./...)
PASS
ok      github.com/chihaya/chihaya/storage/memory       0.005s
PASS
ok      github.com/chihaya/chihaya/storage/memorybysubnet       0.004s
PASS
ok      github.com/chihaya/chihaya/storage/redis        2.675s
```

PS: I have serveral other tests in my test environment and show successfully.

Addition:
* 1. This design doesn't use [redis distributed lock](https://redis.io/topics/distlock) for synchronization, which i think is proper even when multiple Chihaya instances are deployed(there is a need for further discussion).
* 2. I have not used Chihaya with Redis in my production environment yet and the Redis logic itself needs more polish.

Signed-off-by: duyanghao <1294057873@qq.com>

### DIFF
--- a/storage/redis/peer_store.go
+++ b/storage/redis/peer_store.go
@@ -1,0 +1,772 @@
+// Package redis implements the storage interface for a Chihaya
+// BitTorrent tracker keeping peer data in redis.
+package redis
+
+import (
+	"encoding/binary"
+	"net"
+	neturl "net/url"
+	"sync"
+	"time"
+
+	"gopkg.in/yaml.v2"
+
+	"github.com/chihaya/chihaya/bittorrent"
+	"github.com/chihaya/chihaya/pkg/log"
+	"github.com/chihaya/chihaya/pkg/timecache"
+	"github.com/chihaya/chihaya/storage"
+
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/chihaya/chihaya/storage/redis/common"
+	"github.com/garyburd/redigo/redis"
+	"gopkg.in/redsync.v1"
+)
+
+// Name is the name by which this peer store is registered with Chihaya.
+const Name = "redis"
+
+// Default config constants.
+const (
+	defaultPrometheusReportingInterval = time.Second * 1
+	defaultGarbageCollectionInterval   = time.Minute * 3
+	defaultPeerLifetime                = time.Minute * 30
+	defaultRedisBroker                 = "redis://myRedis@127.0.0.1:6379/0"
+)
+
+func init() {
+	// Register the storage driver.
+	storage.RegisterDriver(Name, driver{})
+}
+
+type driver struct{}
+
+func (d driver) NewPeerStore(icfg interface{}) (storage.PeerStore, error) {
+	// Marshal the config back into bytes.
+	bytes, err := yaml.Marshal(icfg)
+	if err != nil {
+		return nil, err
+	}
+
+	// Unmarshal the bytes into the proper config type.
+	var cfg Config
+	err = yaml.Unmarshal(bytes, &cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return New(cfg)
+}
+
+// Config holds the configuration of a redis PeerStore.
+type Config struct {
+	GarbageCollectionInterval   time.Duration `yaml:"gc_interval"`
+	PrometheusReportingInterval time.Duration `yaml:"prometheus_reporting_interval"`
+	PeerLifetime                time.Duration `yaml:"peer_lifetime"`
+	RedisBroker                 string        `yaml:"redis_broker"`
+}
+
+// LogFields renders the current config as a set of Logrus fields.
+func (cfg Config) LogFields() log.Fields {
+	return log.Fields{
+		"name":               Name,
+		"gcInterval":         cfg.GarbageCollectionInterval,
+		"promReportInterval": cfg.PrometheusReportingInterval,
+		"peerLifetime":       cfg.PeerLifetime,
+		"redisBroker":        cfg.RedisBroker,
+	}
+}
+
+// Validate sanity checks values set in a config and returns a new config with
+// default values replacing anything that is invalid.
+//
+// This function warns to the logger when a value is changed.
+func (cfg Config) Validate() Config {
+	validcfg := cfg
+
+	if cfg.RedisBroker == "" {
+		validcfg.RedisBroker = defaultRedisBroker
+		log.Warn("falling back to default configuration", log.Fields{
+			"name":     Name + ".RedisBroker",
+			"provided": cfg.RedisBroker,
+			"default":  validcfg.RedisBroker,
+		})
+	}
+
+	if cfg.GarbageCollectionInterval <= 0 {
+		validcfg.GarbageCollectionInterval = defaultGarbageCollectionInterval
+		log.Warn("falling back to default configuration", log.Fields{
+			"name":     Name + ".GarbageCollectionInterval",
+			"provided": cfg.GarbageCollectionInterval,
+			"default":  validcfg.GarbageCollectionInterval,
+		})
+	}
+
+	if cfg.PrometheusReportingInterval <= 0 {
+		validcfg.PrometheusReportingInterval = defaultPrometheusReportingInterval
+		log.Warn("falling back to default configuration", log.Fields{
+			"name":     Name + ".PrometheusReportingInterval",
+			"provided": cfg.PrometheusReportingInterval,
+			"default":  validcfg.PrometheusReportingInterval,
+		})
+	}
+
+	if cfg.PeerLifetime <= 0 {
+		validcfg.PeerLifetime = defaultPeerLifetime
+		log.Warn("falling back to default configuration", log.Fields{
+			"name":     Name + ".PeerLifetime",
+			"provided": cfg.PeerLifetime,
+			"default":  validcfg.PeerLifetime,
+		})
+	}
+
+	return validcfg
+}
+
+// ParseRedisURL ...
+func ParseRedisURL(url string) (host, password string, db int, err error) {
+	// redis://pwd@host/db
+
+	var u *neturl.URL
+	u, err = neturl.Parse(url)
+	if err != nil {
+		return
+	}
+	if u.Scheme != "redis" {
+		err = errors.New("No redis scheme found")
+		return
+	}
+
+	if u.User != nil {
+		password = u.User.String()
+	}
+
+	host = u.Host
+
+	parts := strings.Split(u.Path, "/")
+	if len(parts) == 1 {
+		db = 0 //default redis db
+	} else {
+		db, err = strconv.Atoi(parts[1])
+		if err != nil {
+			db, err = 0, nil //ignore err here
+		}
+	}
+
+	return
+}
+
+// NewRedisBackend creates RedisBackend instance
+func NewRedisBackend(host, password, socketPath string, db int) *RedisBackend {
+	return &RedisBackend{
+		host:       host,
+		db:         db,
+		password:   password,
+		socketPath: socketPath,
+	}
+}
+
+// open returns or creates instance of Redis connection
+func (rb *RedisBackend) open() redis.Conn {
+	if rb.pool == nil {
+		rb.pool = rb.NewPool(rb.socketPath, rb.host, rb.password, rb.db)
+	}
+	if rb.redsync == nil {
+		var pools = []redsync.Pool{rb.pool}
+		rb.redsync = redsync.New(pools)
+	}
+	return rb.pool.Get()
+}
+
+// New creates a new PeerStore backed by redis.
+func New(provided Config) (storage.PeerStore, error) {
+	cfg := provided.Validate()
+
+	// creates RedisBackend instance
+	h, p, db, err := ParseRedisURL(cfg.RedisBroker)
+	if err != nil {
+		return nil, err
+	}
+
+	ps := &peerStore{
+		cfg:    cfg,
+		rb:     NewRedisBackend(h, p, "", db),
+		closed: make(chan struct{}),
+	}
+
+	// Start a goroutine for garbage collection.
+	ps.wg.Add(1)
+	go func() {
+		defer ps.wg.Done()
+		for {
+			select {
+			case <-ps.closed:
+				return
+			case <-time.After(cfg.GarbageCollectionInterval):
+				before := time.Now().Add(-cfg.PeerLifetime)
+				log.Debug("storage: purging peers with no announces since", log.Fields{"before": before})
+				ps.collectGarbage(before)
+			}
+		}
+	}()
+
+	// Start a goroutine for reporting statistics to Prometheus.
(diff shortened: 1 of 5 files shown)


storage/redis/peer_store.go
```diff
@@ -0,0 +1,787 @@
...
+
+	conn := ps.rb.open()
+	defer conn.Close()
+
+	// Update the peer in the swarm.
+	encoded_ih_S := IPver + "_S_" + ih.String()
+	ct := ps.getClock()
+
+	_, err := conn.Do("HSET", encoded_ih_S, pk, ct)
+
```
mrd0ll4r: These blank lines should be removed at some point, but that's a nitpick for now :)

storage/redis/peer_store.go
```diff
@@ -0,0 +1,787 @@
...
+func (ps *peerStore) getClock() int64 {
+	return timecache.NowUnixNano()
+}
+
+func (ps *peerStore) PutSeeder(ih bittorrent.InfoHash, p bittorrent.Peer) error {
+	IPver := p.IP.AddressFamily.String()
+	log.Debug("storage: PutSeeder", log.Fields{
+		"InfoHash": ih.String(),
+		"Peer":     fmt.Sprintf("[ID: %s, IP: %s(AddressFamily: %s), Port %d]", p.ID.String(), p.IP.String(), IPver, p.Port),
```
mrd0ll4r: This will call `fmt.Sprintf` every time, regardless of whether debug logging is enabled or not - we'll have to figure out something else here, but that's also not a priority rn.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,787 @@
...
+	default:
+	}
+
+	pk := newPeerKey(p)
+
+	conn := ps.rb.open()
+	defer conn.Close()
+
+	// Update the peer in the swarm.
+	encoded_ih_S := IPver + "_S_" + ih.String()
```
mrd0ll4r: This should be `camelCase` at some point

storage/redis/peer_store.go
```diff
@@ -0,0 +1,787 @@
...
+	}
+
+	pk := newPeerKey(p)
+
+	conn := ps.rb.open()
+	defer conn.Close()
+
+	encoded_ih_S := IPver + "_S_" + ih.String()
+
+	DelNum, err := conn.Do("HDEL", encoded_ih_S, pk)
```
mrd0ll4r: This should be `camelCase` at some point

storage/redis/peer_store.go
```diff
@@ -0,0 +1,787 @@
...
+	}
+
+	if seeder {
+		// Append leechers as possible.
+		for _, pk := range con_leechers {
+			if numWant == 0 {
+				break
+			}
+
+			peers = append(peers, decodePeerKey(serializedPeer(pk.([]byte))))
```
mrd0ll4r: Is it guaranteed that these type assertions never blow up or panic?

storage/redis/peer_store.go
```diff
@@ -0,0 +1,787 @@
...
+
+				peers = append(peers, decodePeerKey(serializedPeer(pk.([]byte))))
+				numWant--
+			}
+		}
+	}
+
+	APResult := ""
+	for _, pr := range peers {
+		APResult = fmt.Sprintf("%s Peer:[ID: %s, IP: %s(AddressFamily: %s), Port %d]", APResult, pr.ID.String(), pr.IP.String(), IPver, pr.Port)
```
mrd0ll4r: This, of course, also constructs a big string for each announce, regardless of debug logging being turned on or not

storage/redis/peer_store.go
```diff
@@ -0,0 +1,787 @@
...
+	}
+
+	encoded_ih := ih.String()
+	encoded_ih_L := IPver + "_L_" + encoded_ih // key
+	encoded_ih_S := IPver + "_S_" + encoded_ih // key
+
+	conn := ps.rb.open()
+	defer conn.Close()
+
+	leechers, err := conn.Do("HKEYS", encoded_ih_L)
```
mrd0ll4r: Hypothetically, for a swarm with 50,000 leechers, will this actually retrieve all 50,000 peer keys?

storage/redis/peer_store.go
```diff
@@ -0,0 +1,787 @@
...
+
+	return
+}
+
+// collectGarbage deletes all Peers from the PeerStore which are older than the
+// cutoff time.
+//
+// This function must be able to execute while other methods on this interface
+// are being executed in parallel.
+func (ps *peerStore) collectGarbage(cutoff time.Time) error {
```
mrd0ll4r: Redis has some concept of key expiry, right? I'm not an expert about redis, but that would've been my first choice - the garbage collection is just used for the memory implementation because Go doesn't provide kex expiry. Why should we/should we not use the redis feature?

storage/redis/peer_store.go
```diff
@@ -0,0 +1,787 @@
...
+		for _, ih := range infohashes {
+			ih_str := string(ih.([]byte))
+
+			ih_list, err := conn.Do("HGETALL", ih_str) // field
+			if err != nil {
+				return err
+			}
+			con_ih := ih_list.([]interface{})
+
+			if len(con_ih) == 0 {
```
mrd0ll4r: This looks racy - what if another connection added a peer to the infohash between our `HGETALL` and this check?

storage/redis/peer_store.go
```diff
@@ -0,0 +1,787 @@
...
+			}
+
+			var pk serializedPeer
+			for index, ih_field := range con_ih {
+				if index%2 != 0 { // value
+					mtime, err := strconv.ParseInt(string(ih_field.([]byte)), 10, 64)
+					if err != nil {
+						return err
+					}
+					if mtime <= cutoffUnix {
```
mrd0ll4r: Same question as above - what if someone changed the mtime between the `HGETALL` and this check?

storage/redis/peer_store.go
```diff
@@ -0,0 +1,787 @@
...
+				} else { // key
+					pk = serializedPeer(ih_field.([]byte))
+				}
+			}
+
+			ih_len, err := conn.Do("HLEN", ih_str)
+			if err != nil {
+				return err
+			}
+			if ih_len.(int64) == 0 {
```
mrd0ll4r: Same question - what happens if someone adds a key to `ih_len` between the `HLEN` call and this check?

common/redis.go
```diff
@@ -0,0 +1,61 @@
+package common
```
mrd0ll4r: This should be moved to either `pkg/redis/redis.go` or just `storage/redis/common.go` (or something like that), because nothing else in chihaya uses redis (yet).

mrd0ll4r:
Thanks for the PR!

About redis distributed locks: From what I understand, we could either

- Use redis distributed locks for explicit synchronization. I expect this to have some runtime overhead, but haven't tried it myself yet. Also programming this might be annoying.
- Use redis scripts/transactions. I expect this to be faster than explicit locking, but I have no idea how difficult it is to write these scripts.

I think most issues about concurrency are connected to garbage collection, so the third option would (maybe) be:

- Use redis key expiry to do the garbage collection for us. I feel like this'd be an elegant solution, but again - never tried it myself.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,787 @@
...
+	cfg Config
+	rb  *RedisBackend
+
+	closed chan struct{}
+	wg     sync.WaitGroup
+}
+
+var _ storage.PeerStore = &peerStore{}
+
+// populateProm aggregates metrics over all shards and then posts them to
```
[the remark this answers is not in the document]
pushrax: Sharding at the connection level could allow for implementing a horizontally scalable tracker (that can hold more torrents than the memory of any individual server), but that's such an extremely rare use-case that it's probably not worth supporting. Usually it's easier to run multiple trackers and shard client-side by using multiple announce URLs.

storage/redis/peer_store.go
pushrax: It would, perhaps some tricks with HSCAN could work here.

storage/redis/peer_store.go
pushrax: Redis can't expire within a hash. A hash makes sense here to provide performant iteration, though a hash will also have issues with many keys when there are many peers.

pushrax:
Redis transactions are easy to use (`MULTI` + `EXEC`). Scripts (via Lua) are a bit more complicated, we should probably avoid them but for something like the peer retrieval routine it might make sense.

common/redis.go
```diff
@@ -0,0 +1,61 @@
...
+			c, err := rc.open(socketPath, host, password, db)
+			if err != nil {
+				return nil, err
+			}
+
+			if db != 0 {
+				_, err = c.Do("SELECT", db)
+			}
+
+			if err != nil {
```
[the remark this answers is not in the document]
duyanghao: Done fix : )

common/redis.go
duyanghao: Move common package to `storage/redis/common` : )

storage/redis/peer_store.go
```diff
@@ -0,0 +1,787 @@
...
+		})
+	}
+
+	return validcfg
+}
+
+// ParseRedisURL ...
+func ParseRedisURL(url string) (host, password string, db int, err error) {
+	// redis://pwd@host/db
+	parts := strings.Split(url, "redis://")
```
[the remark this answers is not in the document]
duyanghao: Done Rewrite ParseRedisURL using url.Parse

common/redis.go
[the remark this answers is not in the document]
duyanghao: Done fix : )

storage/redis/peer_store.go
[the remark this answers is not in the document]
duyanghao: @mrd0ll4r Remove this line : )
@pushrax All redis connections are managed through [redis pool](https://github.com/garyburd/redigo/blob/master/redis/pool.go#L171)

storage/redis/peer_store.go
duyanghao: Done Remove blank lines : )

storage/redis/peer_store.go
duyanghao: @mrd0ll4r `log.Debug` will works only when debug logging is enabled.
```
A customizable, multi-protocol BitTorrent Tracker

Usage:
  chihaya [flags]

Flags:
      --config string       location of configuration file (default "/etc/chihaya.yaml")
      --cpuprofile string   location to save a CPU profile
      --debug               enable debug logging
  -h, --help                help for chihaya
      --json                enable json logging
      --nocolors            disable log coloring
```

storage/redis/peer_store.go
mrd0ll4r: Oh, that's unfortunate. Thanks for clearing that up!

Alternative idea: Implement the whole garbage collectio thing as a lua script? Although that would completely stop-the-world while it's running, right?

storage/redis/peer_store.go
duyanghao: Done `camelCase` : )

storage/redis/peer_store.go
duyanghao: Done `camelCase` also : )

storage/redis/peer_store.go
duyanghao: I cannot guarantee, but It should work fine as far as I have learned.

storage/redis/peer_store.go
duyanghao: Yes, but i can't figure out a more useful way to debug.

storage/redis/peer_store.go
duyanghao: @mrd0ll4r yes, redis [HKEYS](https://redis.io/commands/hkeys) will return all relevant leechers(redis hash `field`) of the `HashInfo`(redis hash `key`)
@pushrax you are right, maybe we can use `HSCAN` get Peers in batches after all redis is single thread designed.

storage/redis/peer_store.go
duyanghao: I do think it's ok under this circumstances as `lost one or serveral peers` during this interval doesn't affect the overall Peers conditions(there are still many other peers we can acquire).

storage/redis/peer_store.go
duyanghao: It doesn't affect the overall Peers conditions as above.

storage/redis/peer_store.go
duyanghao: It doesn't affect the overall Peers conditions as above.

storage/redis/peer_store.go
duyanghao: @mrd0ll4r @pushrax Maybe it is unnecessary to do synchronization during garbage collection as `lost one or serveral peers` during this interval doesn't affect the overall Peers conditions(there are still many other peers we can acquire).

duyanghao:
@mrd0ll4r @pushrax thanks for all your comments, there are two main problems for us to solve.
1. Is there any necessity we do synchronization during garbage collection? if yes, how to do?
2. use redis `HSACN` to get Peer in batches instead of one-off `HKEYS`

storage/redis/peer_store.go
mrd0ll4r: That's true, yes. Off the top of my head, the only "important" requirement would be that inactive peers/swarms are cleaned up eventually. Opinions @pushrax @jzelinskie ?

mrd0ll4r:
Maybe @pushrax can chime in here, but these are my opinions:

1. If you manage to pull off correct garbage collection without synchronization, for example with scripts or transactions, go for it! The GC has to be correct in that it never deletes things that shouldn't be deleted. I think it's no big deal if it doesn't always clean up everything perfectly, as long as it eventually does so.

2. `HSCAN` sounds like a good idea to me! Does that lock redis for the time it's running though? (Again, haven't worked with redis a whole lot...)

jzelinskie:
@duyanghao Any updates on progress here? I really would like to replace our `memorybysubnet` backend with this redis-based one.

duyanghao:
> @duyanghao Any updates on progress here? I really would like to replace our `memorybysubnet` backend with this redis-based one.

@imroc Would you like to take over this job?

choury:
@hustcat Are Changhongli  working for this now?

onestraw:
Hi @duyanghao, I am considering using Redis storage in production, I want to ask if there are some reasons that make you discard the PR? Will Redis become the bottleneck?

onestraw:
@duyanghao @mrd0ll4r I will take over the PR, is it OK?

jzelinskie:
I think it's safe to say that #433 obviates this PR.

duyanghao:
> @duyanghao @mrd0ll4r I will take over the PR, is it OK?

Sure! Good job. I'm sorry for replying you so late due to private affairs :)

PR #394, opened 2018-02-13
bittorrent: implement nicer String methods for various types

### LINKED ISSUE #390: Implement proper fmt.Stringer for Infohashes, PeerIDs
Our current 'implementation' just barfs the raw bytes as a string, this leads to mangled output in logs and other places.

We should determine if anything depends on the current behaviour of `String()` on these two types and, if yes, maybe move that to a `RawString()` method or something like that. I think we should implement `String()` to return hex or base56 or whatever bittorrent uses by default.

mrd0ll4r:
Fixes #390 

Also adds `String` and `LogFields` methods to `Peer`s.

### DIFF
--- a/bittorrent/bittorrent_test.go
+++ b/bittorrent/bittorrent_test.go
@@ -1,0 +1,59 @@
+package bittorrent
+
+import (
+	"fmt"
+	"net"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	b        = []byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20}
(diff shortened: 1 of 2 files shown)


bittorrent/bittorrent_test.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+func TestPeerID_String(t *testing.T) {
+	s := PeerIDFromBytes(b).String()
+	require.Equal(t, expected, s)
+}
+
+func TestPeerID_RawString(t *testing.T) {
+	s := PeerIDFromBytes(b).RawString()
+	require.Equal(t, string(b), s)
```
jzelinskie: i don't feel like this is testing much

mrd0ll4r:
ptal

PR #395, opened 2018-02-13
Opentracker v6

### LINKED ISSUE #391: Implement IPv6 UDP announces the way opentracker does
...which is also the way BEP15 describes it now.

The implementation we did is from that ancient [blog post](https://web.archive.org/web/20170503181830/http://opentracker.blog.h3q.com/2007/12/28/the-ipv6-situation/), but opentracker [doesn't actually implement that nowadays](http://erdgeist.org/gitweb/opentracker/tree/ot_udp.c#n137). Specifically, they implement a version which uses action `1`, not `4`, and leaves the announce request structure unchanged. This means the IP field becomes useless because it's 32 bits long.

We could consider implementing both the `action==4` way (peers seem to be using it, judging from stats) and the `action==1` version.

mrd0ll4r:
Fixes #391 

Turns out we almost already did this - just the responses were mangled.
Also updated docs to indicate we can do both now.

The "old" opentracker IPv6 way has the advantage of having a usable `ip` field in the announce, to enable IP spoofing.

### DIFF
--- a/cmd/chihaya-e2e/main.go
+++ b/cmd/chihaya-e2e/main.go
@@ -15,8 +15,8 @@ import (
 )
 
 func init() {
-	flag.StringVar(&httpTrackerURL, "http", "http://localhost:6969/announce", "the address of the HTTP tracker")
-	flag.StringVar(&udpTrackerURL, "udp", "udp://localhost:6969", "the address of the UDP tracker")
+	flag.StringVar(&httpTrackerURL, "http", "http://127.0.0.1:6969/announce", "the address of the HTTP tracker")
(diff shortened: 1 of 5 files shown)


jzelinskie:
Is there a way we can test this? Maybe we need to add code to the e2e tool.

mrd0ll4r:
I'll look into it

mrd0ll4r:
`github.com/anacrolix/torrent` doesn't seem to support IPv6 - neither HTTP nor any of the UDP versions. I could write that by hand, but it's gonna get relatively big.

For now I forced the e2e tool to use IPv4.

PR #408, opened 2018-09-04
http: fix TLS

### LINKED ISSUE #407: http: TLS not used despite proper config
How to configure https, only certificate settings, no port, open

mrd0ll4r:
Fixes #407

### DIFF
--- a/frontend/http/frontend.go
+++ b/frontend/http/frontend.go
@@ -164,8 +164,16 @@ func (f *Frontend) listenAndServe() error {
 	f.srv.SetKeepAlivesEnabled(false)
 
 	// Start the HTTP server.
-	if err := f.srv.ListenAndServe(); err != http.ErrServerClosed {
-		return err
+	if f.tlsCfg != nil {
+		// ... using TLS.
+		if err := f.srv.ListenAndServeTLS("", ""); err != http.ErrServerClosed {
+			return err
+		}
+	} else {
+		// ... using plain TCP.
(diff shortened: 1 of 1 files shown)


frontend/http/frontend.go
```diff
@@ -164,8 +164,16 @@ func (f *Frontend) listenAndServe() error {
 	f.srv.SetKeepAlivesEnabled(false)
 
 	// Start the HTTP server.
-	if err := f.srv.ListenAndServe(); err != http.ErrServerClosed {
-		return err
+	if f.tlsCfg != nil {
+		// ... using TLS.
+		if err := f.srv.ListenAndServeTLS("", ""); err != http.ErrServerClosed {
```
jzelinskie: Does this use our TLS Config? I'm not sure that it does: https://golang.org/src/net/http/server.go?s=94692:94760#L3048

mrd0ll4r: Yup, tried it. From the doc on Server.TLSConfig:
```
        // TLSConfig optionally provides a TLS configuration for use
        // by ServeTLS and ListenAndServeTLS. Note that this value is
        // cloned by ServeTLS and ListenAndServeTLS, so it's not
        // possible to modify the configuration with methods like
        // tls.Config.SetSessionTicketKeys. To use [...]
```

We could just provide the cert and key strings here, but the way we're doing it now we fail earlier, and not in a spun-off goroutine, but rather in the main goroutine that set up the HTTP frontend.

PR #414, opened 2018-09-09
Add https_addr config, required to run HTTPS

### LINKED ISSUE #410: http and https  Simultaneous ？
Some old clients only support http,
So you need http and https,
How to configure?

pushrax:
Closes #410, by allowing both `addr` and `https_addr` to be set.

Should we add HTTPS to the e2e test suite?

This is a user-facing breaking change; TLS users will need to switch from `addr` to `https_addr`.

### DIFF
--- a/example_config.yaml
+++ b/example_config.yaml
@@ -25,8 +25,12 @@ chihaya:
   # If you do not wish to run this, delete this section.
   http:
     # The network interface that will bind to an HTTP server for serving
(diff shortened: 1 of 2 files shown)


jzelinskie:
I think it's fine to add that to the e2e tests in another PR.

pushrax:
Rebased and added the correct call to `SetKeepAlivesEnabled`.

PR #415, opened 2018-09-12
middleware : add torrent approval package

### LINKED ISSUE #375: [Feature]: Black/White List
**Blacklist Mode** 
 Block one or more torrent files identified by something unique (e.g. _infohash_) from being tracked by the server for different reasons (e.g. illegal content) and this list can be stored anywhere (e.g DB/TXT).

**Whitelist Mode** 
 Allow one or more torrent files identified by something unique  (e.g. _infohash_) to be tracked by the server for different reason (e.g. private tracker) and this list can be stored anywhere (e.g DB/TXT).

**Dual Mode** 
 Allow all and when needed block one or more torrent files identified by something unique (e.g. _infohash_) and this list can be stored anywhere (e.g DB/TXT).

PS: the first two modes are inspired by opentracker, unfortunately it can't be activated simultaneously (Dual Mode) in it, and the tracker needs to be restarted everytime the list changed.

daftaupe:
this makes it possible to specify a torrent whitelist/blacklist in the
configuration file based on their hexadecimal hash.

Fixes #375

### DIFF
--- a/middleware/torrentapproval/torrentapproval.go
+++ b/middleware/torrentapproval/torrentapproval.go
@@ -1,0 +1,109 @@
+// Package torrentapproval implements a Hook that fails an Announce based on a
+// whitelist or blacklist of torrent hash.
+package torrentapproval
+
+import (
+	"context"
+	"encoding/hex"
+	"fmt"
+
+	"gopkg.in/yaml.v2"
+
+	"github.com/chihaya/chihaya/bittorrent"
+	"github.com/chihaya/chihaya/middleware"
+)
+
+// Name is the name by which this middleware is registered with Chihaya.
+const Name = "torrent approval"
+
+func init() {
+	middleware.RegisterDriver(Name, driver{})
+}
+
+var _ middleware.Driver = driver{}
+
+type driver struct{}
+
+func (d driver) NewHook(optionBytes []byte) (middleware.Hook, error) {
+	var cfg Config
+	err := yaml.Unmarshal(optionBytes, &cfg)
+	if err != nil {
+		return nil, fmt.Errorf("invalid options for middleware %s: %s", Name, err)
+	}
+
+	return NewHook(cfg)
+}
+
+// ErrTorrentUnapproved is the error returned when a torrent hash is invalid.
+var ErrTorrentUnapproved = bittorrent.ClientError("unapproved torrent")
+
+// Config represents all the values required by this middleware to validate
+// torrents based on their hash value.
+type Config struct {
+	Whitelist []string `yaml:"whitelist"`
+	Blacklist []string `yaml:"blacklist"`
+}
+
+type hook struct {
+	approved   map[bittorrent.InfoHash]struct{}
+	unapproved map[bittorrent.InfoHash]struct{}
+}
+
+// NewHook returns an instance of the torrent approval middleware.
+func NewHook(cfg Config) (middleware.Hook, error) {
+	h := &hook{
+		approved:   make(map[bittorrent.InfoHash]struct{}),
+		unapproved: make(map[bittorrent.InfoHash]struct{}),
+	}
+
+	if len(cfg.Whitelist) > 0 && len(cfg.Blacklist) > 0 {
+		return nil, fmt.Errorf("using both whitelist and blacklist is invalid")
+	}
+
+	for _, hashString := range cfg.Whitelist {
+		hashinfo, err := hex.DecodeString(hashString)
+		if err != nil {
+			return nil, fmt.Errorf("whitelist : invalid hash %s", hashString)
+		}
+		if len(hashinfo) != 20 {
+			return nil, fmt.Errorf("whitelist : hash %s is not 20 byes", hashString)
+		}
+		h.approved[bittorrent.InfoHashFromBytes(hashinfo)] = struct{}{}
+	}
+
+	for _, hashString := range cfg.Blacklist {
+		hashinfo, err := hex.DecodeString(hashString)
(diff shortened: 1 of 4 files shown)


middleware/torrentapproval/torrentapproval.go
```diff
@@ -0,0 +1,106 @@
...
+			return nil, fmt.Errorf("blacklist : hash %s is not 20 byes", hashString)
+		}
+		h.unapproved[bittorrent.InfoHashFromBytes(hashinfo)] = struct{}{}
+	}
+
+	return h, nil
+}
+
+func (h *hook) HandleAnnounce(ctx context.Context, req *bittorrent.AnnounceRequest, resp *bittorrent.AnnounceResponse) (context.Context, error) {
+
```
mrd0ll4r: Remove this empty line

middleware/torrentapproval/torrentapproval.go
```diff
@@ -0,0 +1,106 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("whitelist : invalid hash %s", hashString)
+		}
+		if len(hashinfo) != 20 {
+			return nil, fmt.Errorf("whitelist : hash %s is not 20 byes", hashString)
+		}
+		h.approved[bittorrent.InfoHashFromBytes(hashinfo)] = struct{}{}
+	}
+
+	for _, hashString := range cfg.Blacklist {
```
mrd0ll4r: Can you add a log message if both `cfg.Whitelist` and `cfg.Blacklist` are non-empty? Explain how it is weird to use both black- and whitelist at the same time. You could even return an error in that case tbh - I don't see a use case for blacklist and whitelist at the same time.

jzelinskie:
Hi @daftaupe! Glad to see a new contributor!

If you have some time, I'd appreciate some feedback on the process for you to learn about the project and how to write a new piece of middleware. Hopefully we can incorporate that into an improved documentation or code to make things more approachable.

middleware/torrentapproval/torrentapproval_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// Testing blacklisting
+	// any hash not blacklisted is approved
+	// any hash blacklisted is unapproved
+	h, err = NewHook(Config{
+		[]string{},
+		[]string{"85ffc46155d6e0c6f35c8a6dee972305ce3808fe"},
+	})
+
+	// This hash is blacklisted
```
mrd0ll4r: This hash is not blacklisted :)

middleware/torrentapproval/torrentapproval_test.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"context"
+	"encoding/hex"
+	"testing"
+
+	"github.com/chihaya/chihaya/bittorrent"
+	"github.com/stretchr/testify/require"
+)
+
+func TestHandleAnnounce(t *testing.T) {
```
mrd0ll4r: Nice!
Now we're going to turn this into a table-driven test. That means we'll define a table of test cases and what should happen in each case.
Something like this:

```go
var cases = []struct{
    cfg Config
    ih string
    approved bool
}{
    // Infohash is whitelisted
    {
        cfg: Config {
            Whitelist: []string{"3532cf2d327fad8448c075b4cb42c8136964a435"},
        },
        ih: "3532cf2d327fad8448c075b4cb42c8136964a435",
        approved: true,
    },//or write a shortened version:
    // Infohash is not whitelisted
    { Config{Whitelist: []string{"3532cf2d327fad8448c075b4cb42c8136964a435"}}, "4532cf2d327fad8448c075b4cb42c8136964a435", false},
    { ....}
}
```

And then you can simplify your test by iterating over the cases. It'd be super nice if you used sub-tests with `t.Run(...)`. You can find an example of what all this looks like in [frontend/udp/connection_id_test.go](https://github.com/chihaya/chihaya/blob/master/frontend/udp/connection_id_test.go).

daftaupe:
@mrd0ll4r this should have improved the test section, as you requested.

middleware/torrentapproval/torrentapproval_test.go
```diff
@@ -0,0 +1,73 @@
...
+		},
+		"3532cf2d327fad8448c075b4cb42c8136964a435",
+		false,
+	},
+}
+
+func TestHandleAnnounce(t *testing.T) {
+	for _, tt := range cases {
+		t.Run(fmt.Sprintf("testing hash %s", tt.ih), func(t *testing.T) {
+			h, _ := NewHook(tt.cfg)
```
mrd0ll4r: Check the error with `require.Nil(t,err)`

middleware/torrentapproval/torrentapproval_test.go
```diff
@@ -0,0 +1,73 @@
...
+			ctx := context.Background()
+			req := &bittorrent.AnnounceRequest{}
+			resp := &bittorrent.AnnounceResponse{}
+
+			hashbytes, _ := hex.DecodeString(tt.ih)
```
mrd0ll4r: Same here

middleware/torrentapproval/torrentapproval_test.go
```diff
@@ -0,0 +1,73 @@
...
+			hashinfo := bittorrent.InfoHashFromBytes(hashbytes)
+
+			req.InfoHash = hashinfo
+
+			_, err := h.HandleAnnounce(ctx, req, resp)
```
mrd0ll4r: Check that this context is unchanged with `require.Equal`

PR #423, opened 2018-09-20
middleware/clientapproval : add tests

daftaupe:


daftaupe:
@mrd0ll4r I was planning to add the blacklist/whitelist at the same time in another PR but now it's in this one.

mrd0ll4r:
Well, you're not wrong, it could've been a separate PR. But it's so tiny and you were working on this middleware already, so I figured why not have both in one :)

middleware/clientapproval/clientapproval_test.go
```diff
@@ -0,0 +1,75 @@
...
+			Blacklist: []string{"010203"},
+		},
+		"01020304050607080900",
+		false,
+	},
+}
+
+func TestHandleAnnounce(t *testing.T) {
+	for _, tt := range cases {
+		t.Run(fmt.Sprintf("testing peerid %s", tt.peerID), func(t *testing.T) {
```
mrd0ll4r: Oh, sorry. I just figured out these all have the same name because the peer ID is always the same. The point of naming them, however, is to distinguish the tests and identify which cases are failing. Can you name them in a way that makes it unique and clear which case is being tested?

PR #429, opened 2018-10-19
frontend/http: canonicalize http real_ip_header

### LINKED ISSUE #428: Real IP http header ignore case sensitivity
In order to get the IP address of a client, when using some reverse proxy or load balancer, one can use `real_ip_header` config defined in http section. As http headers are [case insensitive](https://tools.ietf.org/html/rfc2616#section-4.2), was there any reason to not canonicalize header key?

Let me know if I should create a PR to apply following fix.

```diff
diff --git a/frontend/http/parser.go b/frontend/http/parser.go
index 870b0d3..d5e4644 100644
--- a/frontend/http/parser.go
+++ b/frontend/http/parser.go
@@ -153,8 +153,8 @@ func requestedIP(r *http.Request, p bittorrent.Params, opts ParseOptions) (ip ne
 	}
 
 	if opts.RealIPHeader != "" {
-		if ips, ok := r.Header[opts.RealIPHeader]; ok && len(ips) > 0 {
-			return net.ParseIP(ips[0]), false
+		if ip := r.Header.Get(opts.RealIPHeader); ip != "" {
+			return net.ParseIP(ip), false
 		}
 	}
 

```

yashpal1995:
Fixes #428

### DIFF
--- a/frontend/http/parser.go
+++ b/frontend/http/parser.go
@@ -153,8 +153,8 @@ func requestedIP(r *http.Request, p bittorrent.Params, opts ParseOptions) (ip ne
 	}
 
 	if opts.RealIPHeader != "" {
-		if ips, ok := r.Header[opts.RealIPHeader]; ok && len(ips) > 0 {
-			return net.ParseIP(ips[0]), false
+		if ip := r.Header.Get(opts.RealIPHeader); ip != "" {
+			return net.ParseIP(ip), false
 		}
 	}
 


jzelinskie:
The CI failed because of a dependency on our test tooling.
It is unrelated to your change and this should be fine to merge.
I'm going to wait until GitHub isn't having partial outages to see if I can rerunning it passes or if we'll need to make some changes to the CI scripts.

PR #431, opened 2018-10-23
Fix path to golint for travis

pushrax:
Needed for CI to pass.


cmd/chihaya/config.go
```diff
@@ -10,14 +10,10 @@ import (
 	"github.com/chihaya/chihaya/frontend/http"
 	"github.com/chihaya/chihaya/frontend/udp"
 	"github.com/chihaya/chihaya/middleware"
-
-	// Imported to register as middleware drivers.
```
mrd0ll4r: Does this not pass through golint or goimports anymore? That'd be sad, I think it's quite useful here tbh

pushrax: That's correct. I think it's obvious enough from the context, but the other options are:

- put the comment on the end of one of the import lines
- open new `import()` sections

mrd0ll4r: I think we should add more import blocks to keep the comments.

cmd/chihaya/config.go
```diff
@@ -10,14 +10,16 @@ import (
...
+import (
+	// Imports to register middleware drivers.
 	_ "github.com/chihaya/chihaya/middleware/clientapproval"
 	_ "github.com/chihaya/chihaya/middleware/jwt"
 	_ "github.com/chihaya/chihaya/middleware/torrentapproval"
 	_ "github.com/chihaya/chihaya/middleware/varinterval"
-
-	// Imported to register as storage drivers.
+)
+import (
```
jzelinskie: Does gofmt not add new lines between import blocks? If it's true, then LGTM else, I'd prefer them to be here, it looks strange w/o them.

PR #433, opened 2018-12-12
Redis Storage

onestraw:
Based on https://github.com/chihaya/chihaya/pull/389


storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+			"name":     Name + ".PeerLifetime",
+			"provided": cfg.PeerLifetime,
+			"default":  validcfg.PeerLifetime,
+		})
+	}
+
+	return validcfg
+}
+
+// ParseRedisURL parse url like `redis://pwd@host/db`
```
mrd0ll4r: This comment could need a little love

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+
+	host = u.Host
+
+	parts := strings.Split(u.Path, "/")
+	if len(parts) == 1 {
+		db = 0 //default redis db
+	} else {
+		db, err = strconv.Atoi(parts[1])
+		if err != nil {
+			db, err = 0, nil //ignore err here
```
mrd0ll4r: Why ignore this error?

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+		}
+	}
+
+	return
+}
+
+// NewBackend creates Backend instance
```
mrd0ll4r: Needs some love. Whole sentences, with punctuation.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+	} else if len(peer.IP.IP) == net.IPv6len { // implies toReturn.IP.To4() == nil
+		peer.IP.AddressFamily = bittorrent.IPv6
+	} else {
+		panic("IP is neither v4 nor v6")
+	}
+
+	return peer
+}
+
+// Backend represents a Memcache result backend
```
mrd0ll4r: How did Mencache get in here? :)

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+		host:       host,
+		db:         db,
+		password:   password,
+		socketPath: socketPath,
+	}
+}
+
+// open returns or creates instance of Redis connection
+func (rb *Backend) open() redis.Conn {
+	if rb.pool == nil {
```
mrd0ll4r: These checks and assignments could be called concurrently from multiple goroutines, is that a problem?

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+	defer conn.Close()
+
+	for _, shard := range shards {
+		ihList, err := conn.Do("HKEYS", shard) // key
+		if err != nil {
+			return
+		}
+		infohashes := ihList.([]interface{})
+
+		InfohashLPrefix := shard + "_L_"
```
mrd0ll4r: all of these variable names should be camelCase.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+		InfohashSPrefix := shard + "_S_"
+		InfohashPrefixLen := len(InfohashLPrefix)
+		InfohashesMap := make(map[string]bool)
```
mrd0ll4r: This should be a `map[string]struct{}` to save some memory.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+		}
+
+		numInfohashes += uint64(len(InfohashesMap))
+	}
+
+	storage.PromInfohashesCount.Set(float64(numInfohashes))
+	storage.PromSeedersCount.Set(float64(numSeeders))
+	storage.PromLeechersCount.Set(float64(numLeechers))
+
+	log.Debug("storage: populateProm() aggregates metrics over all shards", log.Fields{
```
mrd0ll4r: 1. We are usually interested in how long this takes and
2. The message could need some love

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+
+func (ps *peerStore) getClock() int64 {
+	return timecache.NowUnixNano()
+}
+
+func (ps *peerStore) PutSeeder(ih bittorrent.InfoHash, p bittorrent.Peer) error {
+	IPver := p.IP.AddressFamily.String()
+	log.Debug("storage: PutSeeder", log.Fields{
+		"InfoHash": ih.String(),
+		"Peer":     fmt.Sprintf("[ID: %s, IP: %s(AddressFamily: %s), Port %d]", p.ID.String(), p.IP.String(), IPver, p.Port),
```
mrd0ll4r: `Peer` implements `fmt.Stringer`, just use `p.String()`.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+func (ps *peerStore) populateProm() {
+	var numInfohashes, numSeeders, numLeechers uint64
+
+	shards := [2]string{bittorrent.IPv4.String(), bittorrent.IPv6.String()}
+
+	conn := ps.rb.open()
+	defer conn.Close()
+
+	for _, shard := range shards {
+		ihList, err := conn.Do("HKEYS", shard) // key
```
mrd0ll4r: This entire method needs some more comments.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+		if err != nil {
+			return
```
mrd0ll4r: We should at least log a warning, maybe even an error.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+	closed chan struct{}
+	wg     sync.WaitGroup
+}
+
+// populateProm aggregates metrics over all shards and then posts them to
+// prometheus.
+func (ps *peerStore) populateProm() {
+	var numInfohashes, numSeeders, numLeechers uint64
+
+	shards := [2]string{bittorrent.IPv4.String(), bittorrent.IPv6.String()}
```
mrd0ll4r: Calling these shards is confusing - they are not shards in the same sense as we have in the memory stoarges. This is a slice of the two IP address families. The actual space of infohashes is not sharded.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+		infohashes := ihList.([]interface{})
+
+		InfohashLPrefix := shard + "_L_"
+		InfohashSPrefix := shard + "_S_"
+		InfohashPrefixLen := len(InfohashLPrefix)
+		InfohashesMap := make(map[string]bool)
+
+		for _, ihKey := range infohashes {
+			ihStr := string(ihKey.([]byte))
+			ih := ihStr[InfohashPrefixLen:]
```
mrd0ll4r: This could panic if `len(ihStr) < InfohashPrefixLen`. Additionally, it is questionable to me why you take the first `InfohashPrefixLen` bytes of this string, to then check whether those have the `InfohashSPrefix` or `InfohashLPrefix` as prefix, instead of just comparing them OR doing the prefix check in the first place, without trimming.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+			if strings.HasPrefix(ihStr, InfohashLPrefix) {
+				numLeechers++
+				InfohashesMap[ih] = true
+			} else if strings.HasPrefix(ihStr, InfohashSPrefix) {
+				numSeeders++
+				InfohashesMap[ih] = true
+			} else {
+				log.Error("storage: invalid Redis state", log.Fields{
```
mrd0ll4r: We should add in the message that this happened during GC, so we can figure out which Hkey and Hfield we were going through.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+		InfohashSPrefix := shard + "_S_"
+		InfohashPrefixLen := len(InfohashLPrefix)
+		InfohashesMap := make(map[string]bool)
+
+		for _, ihKey := range infohashes {
+			ihStr := string(ihKey.([]byte))
+			ih := ihStr[InfohashPrefixLen:]
+			if strings.HasPrefix(ihStr, InfohashLPrefix) {
+				numLeechers++
+				InfohashesMap[ih] = true
```
mrd0ll4r: I'm relatively sure this is not what you want to do: `ih` at this point is something like `IPv4_S_`, because it got trimmed above - there's no information about the infohash in there anymore.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+	}
+
+	return nil
+}
+
+func (ps *peerStore) DeleteSeeder(ih bittorrent.InfoHash, p bittorrent.Peer) error {
+	IPver := p.IP.AddressFamily.String()
+	log.Debug("storage: DeleteSeeder", log.Fields{
+		"InfoHash": ih.String(),
+		"Peer":     fmt.Sprintf("[ID: %s, IP: %s(AddressFamily: %s), Port %d]", p.ID.String(), p.IP.String(), IPver, p.Port),
```
mrd0ll4r: Same as above

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+	}
+
+	pk := newPeerKey(p)
+
+	conn := ps.rb.open()
+	defer conn.Close()
+
+	encodedSeederInfoHash := IPver + "_S_" + ih.String()
+
+	DelNum, err := conn.Do("HDEL", encodedSeederInfoHash, pk)
```
mrd0ll4r: Variable names should be camelCase

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+	log.Debug("storage: recordGCDuration", log.Fields{"timeTaken(ms)": float64(duration.Nanoseconds()) / float64(time.Millisecond)})
+	storage.PromGCDurationMilliseconds.Observe(float64(duration.Nanoseconds()) / float64(time.Millisecond))
+}
+
+func (ps *peerStore) getClock() int64 {
+	return timecache.NowUnixNano()
+}
+
+func (ps *peerStore) PutSeeder(ih bittorrent.InfoHash, p bittorrent.Peer) error {
+	IPver := p.IP.AddressFamily.String()
```
mrd0ll4r: Probably better name these `addressFamily` to clear up some confusion.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+	}
+
+	return nil
+}
+
+func (ps *peerStore) PutLeecher(ih bittorrent.InfoHash, p bittorrent.Peer) error {
+	IPver := p.IP.AddressFamily.String()
+	log.Debug("storage: PutLeecher", log.Fields{
+		"InfoHash": ih.String(),
+		"Peer":     fmt.Sprintf("[ID: %s, IP: %s(AddressFamily: %s), Port %d]", p.ID.String(), p.IP.String(), IPver, p.Port),
```
mrd0ll4r: Same as above.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+	}
+
+	return nil
+}
+
+func (ps *peerStore) DeleteLeecher(ih bittorrent.InfoHash, p bittorrent.Peer) error {
+	IPver := p.IP.AddressFamily.String()
+	log.Debug("storage: DeleteLeecher", log.Fields{
+		"InfoHash": ih.String(),
+		"Peer":     fmt.Sprintf("[ID: %s, IP: %s(AddressFamily: %s), Port %d]", p.ID.String(), p.IP.String(), IPver, p.Port),
```
mrd0ll4r: Same as above.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+	}
+
+	return nil
+}
+
+func (ps *peerStore) GraduateLeecher(ih bittorrent.InfoHash, p bittorrent.Peer) error {
+	IPver := p.IP.AddressFamily.String()
+	log.Debug("storage: GraduateLeecher", log.Fields{
+		"InfoHash": ih.String(),
+		"Peer":     fmt.Sprintf("[ID: %s, IP: %s(AddressFamily: %s), Port %d]", p.ID.String(), p.IP.String(), IPver, p.Port),
```
mrd0ll4r: Same as above.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+	return nil
+}
+
+func (ps *peerStore) AnnouncePeers(ih bittorrent.InfoHash, seeder bool, numWant int, announcer bittorrent.Peer) (peers []bittorrent.Peer, err error) {
+	IPver := announcer.IP.AddressFamily.String()
+	log.Debug("storage: AnnouncePeers", log.Fields{
+		"InfoHash": ih.String(),
+		"seeder":   seeder,
+		"numWant":  numWant,
+		"Peer":     fmt.Sprintf("[ID: %s, IP: %s(AddressFamily: %s), Port %d]", announcer.ID.String(), announcer.IP.String(), IPver, announcer.Port),
```
mrd0ll4r: Same as above.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+					break
+				}
+
+				peers = append(peers, decodePeerKey(serializedPeer(pk.([]byte))))
+				numWant--
+			}
+		}
+	}
+
+	APResult := ""
```
mrd0ll4r: What is this entire block for? If you want a string representation of the last peer in the list, use

```go
if len(peers) > 0 {
    lastPeerStr = peers[len(peers)-1].String()
    // use lastPeerStr...
}
```

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+	encodedInfoHash := ih.String()
+	encodedLeecherInfoHash := IPver + "_L_" + encodedInfoHash // key
+	encodedSeederInfoHash := IPver + "_S_" + encodedInfoHash  // key
+
+	conn := ps.rb.open()
+	defer conn.Close()
+
+	leechersLen, err := conn.Do("HLEN", encodedLeecherInfoHash)
+	if err != nil {
+		log.Error("storage: Redis HLEN failure", log.Fields{
```
mrd0ll4r: @jzelinskie We might consider adding an error return value to `ScrapeSwarm` - I don't really remember why we don't have one...

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+// This function must be able to execute while other methods on this interface
+// are being executed in parallel.
+func (ps *peerStore) collectGarbage(cutoff time.Time) error {
+	select {
+	case <-ps.closed:
+		return nil
+	default:
+	}
+
+	shards := [2]string{bittorrent.IPv4.String(), bittorrent.IPv6.String()}
```
mrd0ll4r: Again, not really shards

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+				log.Debug("storage: Deleting Redis", log.Fields{
+					"Hkey":   shard,
+					"Hfield": ihStr,
+				})
+				continue
+			}
+
+			var pk serializedPeer
+			for index, ihField := range conIhList {
+				if index%2 != 0 { // value
```
mrd0ll4r: `if index%2 == 1` is more readable, maybe also add a comment why this needs to be checked in the first place, i.e. what is returned from HGETALL

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+				_, err := conn.Do("DEL", ihStr)
+				if err != nil {
+					return err
+				}
+				log.Debug("storage: Deleting Redis", log.Fields{"Hkey": ihStr})
+				_, err = conn.Do("HDEL", shard, ihStr)
+				if err != nil {
+					return err
+				}
+				log.Debug("storage: Deleting Redis", log.Fields{
```
mrd0ll4r: I hope you're not deleting redis! :o

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
```
mrd0ll4r: These debug messages need some love too.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,803 @@
...
+	}()
+
+	return c.Result()
+}
+
+func (ps *peerStore) LogFields() log.Fields {
+	return ps.cfg.LogFields()
+}
+
+func (ps *peerStore) clean() {
```
mrd0ll4r: Oh, this is an interesting question: Do we really want to clear redis when we shut down? Imagine you have two instances of chihaya running on two machines, now for some reason one machine cannot access the internet anymore, so you do whatever with it, but also close Chihaya. That's the gonna purge all the state in redis and renders the second tracker useless in the process.

I know we should clear our stuff out of redis at some point, but I'm not sure where. Comments @pushrax @jzelinskie ?

storage/redis/peer_store.go
onestraw: I will move this block to storage/redis/common and define a RedisURL struct

storage/redis/peer_store.go
onestraw: @duyanghao did this, capture it first

storage/redis/peer_store.go
onestraw: I think @duyanghao  meant memory cache

storage/redis/peer_store.go
onestraw: Yes, I just found @duyanghao didn't use distributed mutex, and will do this with a pool of distributed mutexes

storage/redis/peer_store.go
onestraw: done

storage/redis/peer_store.go
onestraw: done

storage/redis/peer_store.go
onestraw: done

storage/redis/peer_store.go
onestraw: done

storage/redis/peer_store.go
onestraw: done

storage/redis/peer_store.go
onestraw: done

storage/redis/peer_store.go
mrd0ll4r: That will build a slice of strings on every announce, regardless of whether debug logging is turned on.

You actually don't need to log this at all here. The logic logs the generated announce response if debug logging is turned on, see https://github.com/chihaya/chihaya/blob/master/middleware/logic.go#L62

storage/redis/peer_store.go
onestraw: Maybe `group` is more meaningful

storage/redis/peer_store.go
onestraw: done

storage/redis/peer_store.go
onestraw: done

storage/redis/peer_store.go
onestraw: done

storage/redis/peer_store.go
onestraw: haha

storage/redis/peer_store.go
onestraw: done

storage/redis/peer_store.go
onestraw: done

storage/redis/peer_store.go
onestraw: done

storage/redis/peer_store.go
onestraw: Yes, RWMutex is needed

onestraw: Maybe we could init pool/redsync when create Backend

storage/redis/peer_store.go
onestraw: Maybe we can use a redis counter to record the running chihaya, only the last exiting chihaya can clear the data.

jzelinskie: I don't think any chihaya instance should ever clear redis.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,762 @@
...
+	defer conn.Close()
+
+	// Update the peer in the swarm.
+	encodedLeecherInfoHash := addressFamily + "_L_" + ih.String()
+	ct := ps.getClock()
+	_, err := conn.Do("HSET", encodedLeecherInfoHash, pk, ct)
+	if err != nil {
+		return err
+	}
+	_, err = conn.Do("HSET", addressFamily, encodedLeecherInfoHash, ct)
```
jzelinskie: Why not do both of these in a transaction using a lua script or `MULTI`?

storage/redis/peer_store.go
```diff
@@ -0,0 +1,762 @@
...
+
+	return
+}
+
+// collectGarbage deletes all Peers from the PeerStore which are older than the
+// cutoff time.
+//
+// This function must be able to execute while other methods on this interface
+// are being executed in parallel.
+func (ps *peerStore) collectGarbage(cutoff time.Time) error {
```
jzelinskie: Is there a reason we cannot leverage the `EXPIRE` command to avoid chihaya having any logic around collecting garbage?

storage/redis/peer_store.go
```diff
@@ -0,0 +1,762 @@
...
+			return
+		}
+		infohashes := ihList.([]interface{})
+
+		infohashLPrefix := group + "_L_"
+		infohashSPrefix := group + "_S_"
+		infohashPrefixLen := len(infohashLPrefix)
+		infohashesMap := make(map[string]struct{})
+
+		for _, ihKey := range infohashes {
```
jzelinskie: Is this going to scale? It's probably better see if we can leverage `COUNT` in some way or keep our own statistics in redis and use transactions to update those keys with every change to peers (which could be tricky if we use `EXPIRE`).

storage/redis/common/redis.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+}
+
+// Open a new Redis connection
+func (rc *RedisConnector) open(socketPath, host, password string, db int) (redis.Conn, error) {
+	var opts = []redis.DialOption{
+		redis.DialDatabase(db),
+		redis.DialReadTimeout(15 * time.Second),
+		redis.DialWriteTimeout(15 * time.Second),
+		redis.DialConnectTimeout(15 * time.Second),
```
jzelinskie: These should probably be configurable.

storage/redis/common/redis.go
```diff
@@ -0,0 +1,101 @@
+package common
```
jzelinskie: I think this code should just live in `github.com/chihaya/chihaya/storage/redis`. No other package is actually using this code.

storage/redis/peer_store.go
onestraw: good point

storage/redis/peer_store.go
onestraw: @jzelinskie If we use Redis-A first and swtich to Redis-B, who is responsible for cleaning chihaya data in Redis-A.

storage/redis/common/redis.go
onestraw: done

storage/redis/peer_store.go
onestraw: `EXPIRE` can not set timeout on hash's key-value, although it works for the whole hash table

README.md
```diff
@@ -155,6 +155,10 @@ After all PreHooks have executed, any missing response fields that are required
 PostHooks are asynchronous tasks that occur after a response has been delivered to the client.
 Request data is written to the storage asynchronously in one of these PostHooks.
 
+### HA(High Availability)
+
+Chihaya can achieve high availability by using [redis](https://redis.io/) storage backend.
```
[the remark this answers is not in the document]
onestraw: done

storage/redis/common/redis.go
onestraw: done

storage/redis/peer_store.go
mrd0ll4r: Hmm, we could document that we never clear redis, and document which keys are stored in redis, so the user can clean this up. Or we could add a small utility to do just that.

storage/redis/peer_store.go
onestraw: OK, I will remove it.

storage/redis/peer_store.go
onestraw: Fine, simple but secure

storage/redis/peer_store.go
onestraw: A peer has multiple infohashes, incr/decr redis counter on putting/deleting seeder/leecher is not  exactly what we expect, leecher and seeder set are needed which is more time-consuming

mrd0ll4r: > A peer has multiple infohashes

We collect at least these two metrics in prometheus:
- The total number of infohashes
- The total number of peers across all infohashes

We don't collect the number of _distinct_ peers (i.e. the number of _clients_) across all infohashes, as that'd require tremendous amounts of memory or time. If a user is running a _client_ that is a _peer_ in multiple swarms, that's fine, each peer counts for prometheus.

I like the idea of using a counter in redis, but we'd have to make it so that it's actually correct, i.e. a put with the peer already present does not increase the counter. Not sure how simple that is, I'm not doing redis much. Alternatively, is there not something that allows us to range over all the fields in a hash, but without getting them all from redis in one go? Or does that block something? Again, not an expert.

On another note: Looking at this method, I don't think the calculated numLeechers and numSeeders are correct. Neither do I think the number of infohashes is correct. Let me illustrate what I mean:

As far as I understand, redis looks something like this:
```
- IPv4
  - IPv4_S_<infohash 1>: <modification time>
  - IPv4_L_<infohash 1>: <modification time>
  - IPv4_S_<infohash 2>: <modification time>
- IPv4_S_<infohash 1>
  - <peer 1 key>: <modification time>
  - <peer 2 key>: <modification time>
- IPv4_L_<infohash 1>
  - <peer 3 key>: <modification time>
- IPv4_S_<infohash 2>
  - <peer 3 key>: <modification time>
```
**In this case, prometheus should record two swarms and four peers.**

Now, you iterate over each key under `IPv4` and put them in the `infohashesMap`. You count that map to get the number of infohashes/swarms. This is not correct, as you count an infohash with both seeders and leechers as two.

You then also check whether the prefix is `IPv4_S_` or `IPv4_L_` and, depending on the outcome, count that as one seeder or one leecher. This is not correct, as the swarms may have more than one peer in them.

Instead:
- Get all fields under `IPv4` key, lets call them `prefixedInfohashes`
- Split each `prefixedInfohash` of them into `prefix` and `infohash`
- Collect the `infohash` part in a map (I'd call this `swarms`, to avoid confusion), at the end get the count of the map for the number of swarms
- Ask redis about the number `n` of fields under the `prefixedInfohash` key. If the `prefix` indicates seeders, add `n` to the total number of seeders. Vice versa for leechers.
- Repeat for IPv6

mrd0ll4r:
We're also gonna need a config entry in the example config. And a markdown document in the `docs/storage` directory, explaining how this works, how things are stored, how it's configured, ...

storage/redis/peer_store.go
onestraw: You are right, the method is incorrect.

onestraw: There is a question now, redis is a centralized storage, each chihaya reports prometheus in its own period, which is incorrect.

onestraw: We can use redis command `hset` to check if the peer exist present, but we could not optimize the `swarms` metric. I think we can make a compromise to improve performance, to take `IPv4_infohash_count` as seeder's infohash only, or split the `IPv4_infohash_count` to two metrics `IPv4_infohash_seeder_count` and `IPv4_infohash_leecher_count`.

docs/storage/redis.md
```diff
@@ -0,0 +1,69 @@
...
+    - IPv4_L_<infohash 1>
+      - <peer 3 key>: <modification time>
+    - IPv4_S_<infohash 2>
+      - <peer 3 key>: <modification time>
+
+
+In this case, prometheus will record two swarms, three seeders and one leecher.
+
+So tree keys are used to record the count of swarms, seeders and leechers for each group (IPv4, IPv6).
+<pre>
```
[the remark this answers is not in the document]
onestraw: sure

storage/redis/redis.go
```diff
@@ -0,0 +1,135 @@
...
+
+// parseRedisURL parse rawurl into redisURL
+func parseRedisURL(target string) (*redisURL, error) {
+	var u *url.URL
+	u, err := url.Parse(target)
+	if err != nil {
+		return nil, err
+	}
+	if u.Scheme != "redis" {
+		return nil, errors.New("No redis scheme found")
```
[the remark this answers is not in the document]
onestraw: done

storage/redis/peer_store.go
```diff
@@ -0,0 +1,777 @@
...
+	}()
+
+	return c.Result()
+}
+
+func (ps *peerStore) LogFields() log.Fields {
+	return ps.cfg.LogFields()
+}
+
+func (ps *peerStore) groups() []string {
```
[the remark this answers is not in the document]
onestraw: Sure

storage/redis/peer_store.go
```diff
@@ -0,0 +1,777 @@
...
+				})
+				_, err = conn.Do("HDEL", group, ihStr)
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
+
+	{
```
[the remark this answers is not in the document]
onestraw: I am not sure.
Removed.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,777 @@
...
+	conn.Send("MULTI")
+	conn.Send("HSET", encodedSeederInfoHash, pk, ct)
+	conn.Send("HSET", addressFamily, encodedSeederInfoHash, ct)
+	reply, err := conn.Do("EXEC")
+	if err != nil {
+		return err
+	}
+
+	result := reply.([]interface{})
+	if len(result) == 2 {
```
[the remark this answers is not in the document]
onestraw: Yes, it happens when `EXEC` failed. 
I use `redis.Values` to handle this case, see updates later

storage/redis/peer_store.go
```diff
@@ -0,0 +1,777 @@
...
+
+			var pk serializedPeer
+			for index, ihField := range conIhList {
+				if index%2 == 1 { // value
+					mtime, err := strconv.ParseInt(string(ihField.([]byte)), 10, 64)
+					if err != nil {
+						return err
+					}
+					if mtime <= cutoffUnix {
+						_, err := conn.Do("HDEL", ihStr, pk)
```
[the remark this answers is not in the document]
onestraw: Yes, we should check

storage/redis/peer_store.go
```diff
@@ -0,0 +1,777 @@
...
+
+func (ps *peerStore) getClock() int64 {
+	return timecache.NowUnixNano()
+}
+
+func (ps *peerStore) PutSeeder(ih bittorrent.InfoHash, p bittorrent.Peer) error {
+	addressFamily := p.IP.AddressFamily.String()
+	log.Debug("storage: PutSeeder", log.Fields{
+		"InfoHash": ih.String(),
+		"Peer":     p.String(),
```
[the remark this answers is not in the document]
onestraw: It works.

`DEBU[0018] storage: AnnouncePeers                        InfoHash=631295fc22a0fbf092a85a5945ca198dd7599713 Peer="0102030405060708090a0b0c0d0e0f1011121315@[127.0.0.1]:10002" numWant=50 seeder=false`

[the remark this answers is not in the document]
mrd0ll4r: Thanks! That avoids calling `String` when debug loggin is disabled :)

onestraw:
@mrd0ll4r @jzelinskie Is it OK?

mrd0ll4r:
Oh, sorry for the late reply, it's because of Christmas and NYE and all that. I'll take a look today! :)

storage/redis/peer_store_test.go
```diff
@@ -0,0 +1,54 @@
...
+	s "github.com/chihaya/chihaya/storage"
+)
+
+func createNew() s.PeerStore {
+	rs, err := miniredis.Run()
+	if err != nil {
+		panic(err)
+	}
+	redisURL := fmt.Sprintf("redis://@%s/0", rs.Addr())
+	ps, err := New(Config{GarbageCollectionInterval: 10 * time.Minute, PrometheusReportingInterval: 10 * time.Minute, PeerLifetime: 30 * time.Minute, RedisBroker: redisURL})
```
mrd0ll4r: Please add the other config fields to this, it is printing warning messages during the tests.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,773 @@
...
+	default:
+	}
+
+	pk := newPeerKey(p)
+
+	conn := ps.rb.open()
+	defer conn.Close()
+
+	// Update the peer in the swarm.
+	encodedSeederInfoHash := ps.seederInfohashKey(addressFamily, ih.String())
```
mrd0ll4r: You can move these two lines above the `conn := ps.rb.open()` to have the connection for a shorter period of time. Not sure if that's worth anything though...

storage/redis/peer_store.go
```diff
@@ -0,0 +1,773 @@
...
+					if mtime <= cutoffUnix {
+						ret, err := redis.Int64(conn.Do("HDEL", ihStr, pk))
+						if err != nil {
+							return err
+						}
+						// Maybe this peer has been removed since we got the list above with `HGETALL`.
+						if ret == 0 {
+							continue
+						}
+						// DECR seeder/leecher counter
```
mrd0ll4r: Performance optimization: keep track of how many peers you removed, then only decrement once at the end of the infohash loop.

storage/redis/peer_store_test.go
onestraw: done

storage/redis/peer_store.go
```diff
@@ -0,0 +1,773 @@
...
+			}
+			if _, err := conn.Do("DECRBY", decrCounter, removedPeerCount); err != nil {
+				return err
+			}
+
+			ihLen, err := conn.Do("HLEN", ihStr)
+			if err != nil {
+				return err
+			}
+			if ihLen.(int64) == 0 {
```
mrd0ll4r: This is racy - if someone added a new peer to the infohash `ihStr` _after_ we performed `HLEN` but _before_ we execute `DEL`, we will delete this valid peer. Can we make this atomic?

Also I think we need to decrement the infohash counter if we do delete it.

onestraw: @mrd0ll4r I found one way https://redis.io/topics/transactions and fixed it.

mrd0ll4r: Nice! That's pretty much exactly what I had in mind. Didn't know about the WATCH feature, that's pretty nice.

Do we have to wrap the WATCH ... MULTI ... EXEC in a loop in case EXEC fails, i.e. the WATCHed key was modified?

onestraw: Yes, we have to, but we can ignore `EXEC` error as it isn't that critical.

mrd0ll4r: I think you're right, but the reason why is quite complex. I'll write this down here, for future reference:

- The `Delete(Seeder|Leecher)` and `GraduateLeecher` methods never delete an infohash _key_ from an addressFamily _hash_. They also never decrement the infohash counter.
- The `Put(Seeder|Leecher)` and `GraduateLeecher` methods only ever add infohash keys to addressFamily hashes and increment the infohash counter.
- The only method that deletes from the addressFamily hashes is `collectGarbage`, which also decrements the counters. That means that, even if a `Delete(Seeder|Leecher)` call removes the last peer from a swarm, the infohash counter is not changed and the infohash is left in the addressFamily hash until it will be cleaned up by `collectGarbage`.
- `collectGarbage` must run regularly.
- A `WATCH ... MULTI ... EXEC` block fails, if between the `WATCH` and the 'EXEC` any of the watched keys have changed. The location of the `MULTI` doesn't matter.

We have to analyze four cases to prove our algorithm works. I'll characterize them by a tuple (`number of peers in a swarm before WATCH`, `number of peers in the swarm during the transaction`).
1. (0,0), the easy case: The swarm is empty, we watch the key, we execute `HLEN` and find it empty. We remove it and decrement the counter. It stays empty the entire time, the transaction goes through. :heavy_check_mark: 
2. (1,n > 0): The swarm is not empty, we watch the key, we find it non-empty, we unwatch the key. All good. No transaction is made, no transaction fails. :heavy_check_mark: 
3. (0,1): We have to analyze this in two ways.
    - If the change happens before the `HLEN` call, we will see that the swarm is not empty and start no transaction. :heavy_check_mark: 
    - If the change happens after the `HLEN`, we will attempt a transaction and it will fail. This is okay, the swarm is not empty, we will try cleaning it up next time `collectGarbage` runs. :heavy_check_mark: 
4. (1,0): Again, two ways:
    - If the change happens before the `HLEN`, we will see an empty swarm. This situation happens if a call to `Delete(Seeder|Leecher)` removed the last peer asynchronously. We will attempt a transaction, but the transaction will fail. This is okay, the infohash _key_ will remain in the addressFamily _hash_, we will attempt to clean it up the next time 'collectGarbage` runs. :heavy_check_mark: 
    - If the change happens after the `HLEN`, we will not even attempt to make the transaction. The infohash _key_  will remain in the addressFamil _hash_ and we'll attempt to clean it up the next time `collectGarbage` runs. :heavy_check_mark: 

This really took me a while to figure out, and it's not trivial to prove. We should maybe note this down somewhere in the code.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,784 @@
...
+				// Empty hashes are not shown among existing keys,
+				// in other words, it's removed automatically after `HDEL` the last field.
+				//_, err := conn.Do("DEL", ihStr)
+
+				conn.Send("MULTI")
+				conn.Send("HDEL", group, ihStr)
+				conn.Send("DECR", ps.infohashCountKey(group))
```
mrd0ll4r: We must only decrement this counter if the swarm was a seeder swarm. We only increment it in `PutSeeder`.

storage/redis/peer_store.go
onestraw: It's great. How about putting this as `coolectGarbage` comments?

storage/redis/peer_store.go
onestraw: done

storage/redis/peer_store.go
mrd0ll4r: Yes, that sounds good, thank you :)

storage/redis/peer_store.go
```diff
@@ -0,0 +1,826 @@
...
+	pk := newPeerKey(p)
+	ct := ps.getClock()
+
+	conn := ps.rb.open()
+	defer conn.Close()
+
+	conn.Send("MULTI")
+	conn.Send("HSET", encodedLeecherInfoHash, pk, ct)
+	conn.Send("HSET", addressFamily, encodedLeecherInfoHash, ct)
+	reply, err := redis.Values(conn.Do("EXEC"))
```
mrd0ll4r: This fails sometimes. I got it to fail by force re-checking the torrent in Deluge. What this does is it sends a `stopped` event to the tracker, then re-checks, then sends a `started` event to the tracker. This is the log:
```
DEBU[0318] sanitized announce                            1.defaultNumWant=50 1.maxNumWant=100 compact=true downloaded=0 event=stopped eventProvided=true infoHash=c881fda6e95a47581d79c065b70faa1b5de7e99c ipProvided=false left=2799 numWant=0 numWantProvided=true params="&{/announce info_hash=%c8%81%fd%a6%e9ZGX%1dy%c0e%b7%0f%aa%1b%5d%e7%e9%9c&peer_id=-DE13D0-Z*YkFvQ~fnkP&port=60695&uploaded=0&downloaded=0&left=2799&corrupt=0&key=53A818CA&event=stopped&numwant=0&compact=1&no_peer_id=1&supportcrypto=1&redundant=0 map[downloaded:0 left:2799 event:stopped numwant:0 compact:1 supportcrypto:1 redundant:0 peer_id:-DE13D0-Z*YkFvQ~fnkP port:60695 uploaded:0 corrupt:0 key:53A818CA no_peer_id:1] [[200 129 253 166 233 90 71 88 29 121 192 101 183 15 170 27 93 231 233 156]]}" peer="2d4445313344302d5a2a596b4676517e666e6b50@[127.0.0.1]:60695" uploaded=0
2019/01/20 15:02:29 redis.Do(PING) -> ("PONG", <nil>)
2019/01/20 15:02:29 redis.Do(HLEN, "IPv4_L_c881fda6e95a47581d79c065b"...) -> (1, <nil>)
2019/01/20 15:02:29 redis.Do(HLEN, "IPv4_S_c881fda6e95a47581d79c065b"...) -> (0, <nil>)
2019/01/20 15:02:29 redis.Do() -> (<nil>, <nil>)
DEBU[0318] storage: AnnouncePeers                        InfoHash=c881fda6e95a47581d79c065b70faa1b5de7e99c Peer="2d4445313344302d5a2a596b4676517e666e6b50@[127.0.0.1]:60695" numWant=0 seeder=false
2019/01/20 15:02:29 redis.Do(HKEYS, "IPv4_L_c881fda6e95a47581d79c065b"...) -> (["-DE13D0-Z*YkFvQ~fnkP\xed\x17\u007f\x00\x00\x01"], <nil>)
2019/01/20 15:02:29 redis.Do(HKEYS, "IPv4_S_c881fda6e95a47581d79c065b"...) -> ([], <nil>)
2019/01/20 15:02:29 redis.Do() -> (<nil>, <nil>)
DEBU[0318] generated announce response                   compact=true complete=0 interval=30m0s ipv4Peers="[2d4445313344302d5a2a596b4676517e666e6b50@[127.0.0.1]:60695]" ipv6Peers="[]" minInterval=15m0s
DEBU[0318] storage: DeleteSeeder                         InfoHash=c881fda6e95a47581d79c065b70faa1b5de7e99c Peer="2d4445313344302d5a2a596b4676517e666e6b50@[127.0.0.1]:60695"
2019/01/20 15:02:29 redis.Do(HDEL, "IPv4_S_c881fda6e95a47581d79c065b"..., -DE13D0-Z*YkFvQ~fnkP�) -> (0, <nil>)
2019/01/20 15:02:29 redis.Do() -> (<nil>, <nil>)
DEBU[0318] storage: DeleteLeecher                        InfoHash=c881fda6e95a47581d79c065b70faa1b5de7e99c Peer="2d4445313344302d5a2a596b4676517e666e6b50@[127.0.0.1]:60695"
2019/01/20 15:02:29 redis.Do(HDEL, "IPv4_L_c881fda6e95a47581d79c065b"..., -DE13D0-Z*YkFvQ~fnkP�) -> (1, <nil>)
2019/01/20 15:02:29 redis.Do(DECR, "IPv4_L_count") -> (0, <nil>)
2019/01/20 15:02:29 redis.Do() -> (<nil>, <nil>)
DEBU[0318] sanitized announce                            1.defaultNumWant=50 1.maxNumWant=100 compact=true downloaded=0 event=started eventProvided=true infoHash=c881fda6e95a47581d79c065b70faa1b5de7e99c ipProvided=false left=2799 numWant=100 numWantProvided=true params="&{/announce info_hash=%c8%81%fd%a6%e9ZGX%1dy%c0e%b7%0f%aa%1b%5d%e7%e9%9c&peer_id=-DE13D0-Z*YkFvQ~fnkP&port=60695&uploaded=0&downloaded=0&left=2799&corrupt=0&key=53A818CA&event=started&numwant=200&compact=1&no_peer_id=1&supportcrypto=1&redundant=0 map[downloaded:0 left:2799 key:53A818CA numwant:200 supportcrypto:1 port:60695 uploaded:0 event:started compact:1 no_peer_id:1 redundant:0 peer_id:-DE13D0-Z*YkFvQ~fnkP corrupt:0] [[200 129 253 166 233 90 71 88 29 121 192 101 183 15 170 27 93 231 233 156]]}" peer="2d4445313344302d5a2a596b4676517e666e6b50@[127.0.0.1]:60695" uploaded=0
2019/01/20 15:02:30 redis.Do(HLEN, "IPv4_L_c881fda6e95a47581d79c065b"...) -> (0, <nil>)
2019/01/20 15:02:30 redis.Do(HLEN, "IPv4_S_c881fda6e95a47581d79c065b"...) -> (0, <nil>)
2019/01/20 15:02:30 redis.Do() -> (<nil>, <nil>)
DEBU[0318] storage: AnnouncePeers                        InfoHash=c881fda6e95a47581d79c065b70faa1b5de7e99c Peer="2d4445313344302d5a2a596b4676517e666e6b50@[127.0.0.1]:60695" numWant=100 seeder=false
2019/01/20 15:02:30 redis.Do(HKEYS, "IPv4_L_c881fda6e95a47581d79c065b"...) -> ([], <nil>)
2019/01/20 15:02:30 redis.Do(HKEYS, "IPv4_S_c881fda6e95a47581d79c065b"...) -> ([], <nil>)
2019/01/20 15:02:30 redis.Do() -> (<nil>, <nil>)
DEBU[0318] generated announce response                   compact=true complete=0 interval=30m0s ipv4Peers="[2d4445313344302d5a2a596b4676517e666e6b50@[127.0.0.1]:60695]" ipv6Peers="[]" minInterval=15m0s
DEBU[0318] storage: PutLeecher                           InfoHash=c881fda6e95a47581d79c065b70faa1b5de7e99c Peer="2d4445313344302d5a2a596b4676517e666e6b50@[127.0.0.1]:60695"
2019/01/20 15:02:30 redis.Send(MULTI) -> (<nil>)
2019/01/20 15:02:30 redis.Send(HSET, "IPv4_L_c881fda6e95a47581d79c065b"..., -DE13D0-Z*YkFvQ~fnkP�, 1547985749810944415) -> (<nil>)
2019/01/20 15:02:30 redis.Send(HSET, "IPv4", "IPv4_L_c881fda6e95a47581d79c065b"..., 1547985749810944415) -> (<nil>)
2019/01/20 15:02:30 redis.Do(EXEC) -> (<nil>, <nil>)
2019/01/20 15:02:30 redis.Do() -> (<nil>, <nil>)
ERRO[0318] post-announce hooks failed                    error="redigo: nil returned"
```
I don't know why it fails, but it does.

It does not fail all the time.

EDIT: I think I figured this one out. It's because of the UNWATCH failing (because it has the wrong syntax). The key is still WATCHed and the EXEC here fails. That's actually kind of scary, because we WATCHed it in a different connection! I assumed connections were isolated with regard to WATCH MULTI EXEC transactions.

onestraw:
@mrd0ll4r Sorry，I just ran tests and `e2e`

storage/redis/peer_store.go
```diff
@@ -0,0 +1,826 @@
...
+	defer conn.Close()
+
+	for _, group := range ps.groups() {
+		if n, err := conn.Do("GET", ps.infohashCountKey(group)); err != nil {
+			log.Error("storage: GET counter failure", log.Fields{
+				"key":   ps.infohashCountKey(group),
+				"error": err,
+			})
+		} else {
+			numInfohashes += n.(int64)
```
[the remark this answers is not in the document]
onestraw: I modified all type inflection to `redis reply` helpers. 
I will consider testing later

storage/redis/peer_store.go
```diff
@@ -0,0 +1,826 @@
...
+					}
+					if mtime <= cutoffUnix {
+						ret, err := redis.Int64(conn.Do("HDEL", ihStr, pk))
+						if err != nil {
+							return err
+						}
+
+						removedPeerCount += ret
+
+						log.Debug("storage: deleting peer", log.Fields{
```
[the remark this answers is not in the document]
onestraw: done

storage/redis/peer_store.go
```diff
@@ -0,0 +1,826 @@
...
+				_, err = redis.Values(conn.Do("EXEC"))
+				if err != nil {
+					log.Error("storage: Redis EXEC failure, maybe caused by WATCH, ignored", log.Fields{
+						"group":    group,
+						"infohash": ihStr,
+						"error":    err,
+					})
+				}
+			} else {
+				conn.Do("UNWATCH", ihStr)
```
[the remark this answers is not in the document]
onestraw: It's my fault

storage/redis/peer_store.go
mrd0ll4r: This actually means that we have to wrap every `MULTI` ... `EXEC` block in a loop, because it could be failing from an unrelated 'WATCH` done in another connection.

onestraw: @mrd0ll4r Transaction is limited inside connection, WATCH/UNWATCH works in one connection, it doesn't have effects on other connections, I have experimented this by opening two `redis-cli`, the number represents the executing sequence

client-A
```
1. set name onestraw
2. watch name
6. multi
7. get name
8. set name larry.he
9. exec
(nil)
```
client-B
```
3. unwatch
4. set name one-starw
5. get name
```

mrd0ll4r: I agree, that was a mistake on my end. I think I saw the behavior because my requests were so slow and far apart that we simply recycled the same connection, and that still had the WATCH going. So yes, you're right, transactions are limited inside connections. :)

onestraw:
@mrd0ll4r Thank you, I learned a lot from your comments.

example_config.yaml
```diff
@@ -127,6 +127,18 @@ chihaya:
       # are collected and posted to Prometheus.
       prometheus_reporting_interval: 1s
 
+  # This block defines configuration used for redis storage.
```
mrd0ll4r: Ah, sorry, please don't strip the comments, they're probably useful :)

storage/redis/peer_store.go
```diff
@@ -0,0 +1,828 @@
...
+}
+
+func (ps *peerStore) Stop() stop.Result {
+	c := make(stop.Channel)
+	go func() {
+		close(ps.closed)
+		ps.wg.Wait()
+		// chihaya does not clear data in redis when exiting.
+		// chihaya keys have prefix `IPv{4,6}_`.
+		close(c)
```
mrd0ll4r: Please use `c.Done` instead.

storage/redis/peer_store.go
```diff
@@ -0,0 +1,828 @@
...
+	return nil
+}
+
+func (ps *peerStore) Stop() stop.Result {
+	c := make(stop.Channel)
+	go func() {
+		close(ps.closed)
+		ps.wg.Wait()
+		// chihaya does not clear data in redis when exiting.
+		// chihaya keys have prefix `IPv{4,6}_`.
```
mrd0ll4r: Can you turn this into an info-level log message?

onestraw: sure

storage/redis/peer_store.go
onestraw: done

pushrax:
I plan to take a deep look at this in the next week.

storage/redis/peer_store.go
```diff
@@ -815,9 +815,8 @@ func (ps *peerStore) Stop() stop.Result {
 	go func() {
 		close(ps.closed)
 		ps.wg.Wait()
-		// chihaya does not clear data in redis when exiting.
-		// chihaya keys have prefix `IPv{4,6}_`.
-		close(c)
+		log.Info("exiting. chihaya does not clear data in redis when exiting. chihaya keys have prefix 'IPv{4,6}_'.")
```
mrd0ll4r: I'm so sorry to bother you again with this. Please add the `storage: ` prefix to the logged message so people know where it's coming from :)

docs/storage/redis.md
```diff
@@ -0,0 +1,72 @@
+# Redis Storage
+
+This storage system separates chihaya from storage and stores all peer data in Redis to achieve HA.
```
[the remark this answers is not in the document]
onestraw: You're right, it's too brief

jzelinskie:
@pushrax do you still want to review this?

jzelinskie:
I'm going to move forward and merge this. It's better to have something than nothing!

Issue #440, opened 2018-12-27
Enable race detector on travis


mrd0ll4r:
... and fail the build if it finds something?

jzelinskie:
Fixed in #441

PR #438, opened 2018-12-25
Fix 437

cenkalti:
Fix #437

PR #441, opened 2018-12-29
enable -race flag in tests

cenkalti:


cenkalti:
Fixes #440.

Discussed in #438.

PR #442, opened 2018-12-30
split listen&serve for http

### LINKED ISSUE #439: Split listen and serve for HTTP too?
#438 splits listen and serve into two functions for the UDP frontend. This is nice, because instead of just logging an error but otherwise continuing to run, we now actually register an error from `listen` and can handle it, or exit.

We could do the same for HTTP. Just need to be a bit careful about cleaning up, e.g. if listening on the HTTP endpoint works, but listening on the HTTPS endpoint fails, we should probably close the socket we opened earlier.

Opinions?

cenkalti:
Fix #439


mrd0ll4r:
I'm sorry, I know I took a look at this before, but it seems I forgot about it. Thanks a lot for the PR!

PR #448, opened 2019-02-13
vendor: move to Go modules for dependencies

### LINKED ISSUE #444: Update dependency: anacrolix/torrent
The latest [`Announce`](https://github.com/anacrolix/torrent/blob/master/tracker/tracker.go#L53) struct has discarded `HttpClient` field.

```
cmd/chihaya/e2e.go:94:3: unknown field 'HttpClient' in struct literal of type tracker.Announce
cmd/chihaya/e2e.go:122:3: unknown field 'HttpClient' in struct literal of type tracker.Announce
```
https://travis-ci.org/chihaya/chihaya/jobs/474766594

jzelinskie:
Fixes #444.


go.mod
```diff
@@ -0,0 +1,22 @@
...
+	github.com/mendsley/gojwk v0.0.0-20141217222730-4d5ec6e58103
+	github.com/minio/sha256-simd v0.0.0-20190131020904-2d45a736cd16
+	github.com/pkg/errors v0.8.1
+	github.com/prometheus/client_golang v0.9.2
+	github.com/sirupsen/logrus v1.3.0
+	github.com/spf13/cobra v0.0.3
+	github.com/spf13/pflag v1.0.3 // indirect
+	github.com/stretchr/testify v1.3.0
+	github.com/yuin/gopher-lua v0.0.0-20190206043414-8bfc7677f583 // indirect
+	gopkg.in/yaml.v2 v2.2.2
```
mrd0ll4r: You replaced the `redsync` dependency (which was also on gopkg.in) with the original from github, should we do the same for this one or not?

jzelinskie: While I don't like using `gopkg.in` these days, I don't think we should change this one. I believe this package has a magic import comment that says you have to load it from this URL. I had to move the other dependency to GitHub because it wasn't able to to be resolved for some strange reason.

go.mod
```diff
@@ -0,0 +1,22 @@
+module github.com/chihaya/chihaya
+
+require (
+	github.com/SermoDigital/jose v0.0.0-20180104203859-803625baeddc
+	github.com/alicebob/gopher-json v0.0.0-20180125190556-5a6b3ba71ee6 // indirect
+	github.com/alicebob/miniredis v2.4.6+incompatible
+	github.com/anacrolix/torrent v1.0.0
```
anacrolix: I would have expected anacrolix/missinggo and anacrolix/dht to appear in your indirect list. Unless it's because anacrolix/torrent has a go mod file?

jzelinskie: That very well could be. I just blew everything away and regenerated a new go.mod and the only difference is one thing that I hand edited with the jose library, which is because they don't use semver for their git tags.

anacrolix: Meh, it works 👍

mrd0ll4r: > Unless it's because anacrolix/torrent has a go mod file?

That's exactly why :) See [here](https://github.com/golang/go/wiki/Modules#why-does-go-mod-tidy-record-indirect-and-test-dependencies-in-my-gomod), 4th paragraph:

> If a dependency of your module does not itself have a go.mod (e.g., because the dependency has not yet opted in to modules itself), or if its go.mod file is missing one or more of its dependencies (e.g., because the module author did not run go mod tidy), then the missing transitive dependencies will be added to your module's requirements, along with an // indirect comment to indicate that the dependency is not from a direct import within your module.

PR #449, opened 2019-02-14
storage: remove memorybysubnet

jzelinskie:
This code, while interesting, was only relevant to Facebook and they use
their own fork that they maintain. There was not enough outside
interest to warrant maintaining. I'd rather us use the effort to support
a redis storage backend instead.


pushrax:
Why did they need to run a fork?

jzelinskie:
I believe they had patched an older version enough that it would be a huge burden to rebase.
I remember one patch being so that they could run TLS and non-TLS on the same HTTP port, which has a performance overhead.

I want us to be able to support them, but they've never used the code in this PR and neither has anyone else.

cc @cooperlees who would have more details

cooperlees:
This totally kills any chance of us ever coming to v2. I communicated all our reasons when you did the rewrite and made things more "Go" mainline or what ever terminology was used. You broke so much of what we need, it was not worth our time to move to v2.

In regards to subnet matching, I still argue that handing peers back to the same **/24** or same **/48** in IPv6 will be better than random, IF, you have a big enough swarm tho. There are generic carrier boundaries on the Internet that help make this usable.

Other Reasons we didn't move:
- Total config change for no benefit visible benefit to us
- No subnet support (totally now)
- Our HTTPS does 10000's of requests per second - So I guess we can live with "the performance overhead" you speak of ..
- Gigs of memory is fine for us
- What ever morpheous or how ever you do the stats now - made my life harder - so was another reason to not move.

Thanks for the original project, it's still humming away for us ([source here](https://github.com/psaab/chihaya)). I hope v2 is working well where you have all deployed it.

PR #459, opened 2019-10-12
Frontend defaults

### LINKED ISSUE #443: No defaults for frontend parser options
When using with empty config, because `MaxNumWant` is zero by default, no peers are returned to announce requests. Sensible defaults should be set for these config variables.

https://github.com/chihaya/chihaya/blob/b4b257c1515b6efdec22cb22526d2d421f3f9098/frontend/http/parser.go#L18-L20

https://github.com/chihaya/chihaya/blob/b4b257c1515b6efdec22cb22526d2d421f3f9098/frontend/udp/parser.go#L56-L58

mrd0ll4r:
Fixes #443

Probably doesn't build yet because of the formatting of `cmd/chihaya/config.go`.

### DIFF
--- a/frontend/http/parser.go
+++ b/frontend/http/parser.go
@@ -20,6 +20,13 @@ type ParseOptions struct {
 	MaxScrapeInfoHashes uint32 `yaml:"max_scrape_infohashes"`
 }
 
+// Default parser config constants.
+const (
+	defaultMaxNumWant          uint32 = 100
+	defaultDefaultNumWant      uint32 = 50
+	defaultMaxScrapeInfoHashes uint32 = 50
+)
+
 // ParseAnnounce parses an bittorrent.AnnounceRequest from an http.Request.
 func ParseAnnounce(r *http.Request, opts ParseOptions) (*bittorrent.AnnounceRequest, error) {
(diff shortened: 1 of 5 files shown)


frontend/http/parser.go
```diff
@@ -20,6 +20,13 @@ type ParseOptions struct {
 	MaxScrapeInfoHashes uint32 `yaml:"max_scrape_infohashes"`
 }
 
+// Default parser config constants.
+const (
+	defaultMaxNumWant          uint32 = 100
+	defaultDefaultNumWant      uint32 = 50
+	defaultMaxScrapeInfoHashes uint32 = 50
```
jzelinskie: You can leave the type off the const. Constant numbers can actually be "untyped" in Go. It's slightly more flexible.

frontend/udp/parser.go
```diff
@@ -58,6 +58,13 @@ type ParseOptions struct {
 	MaxScrapeInfoHashes uint32 `yaml:"max_scrape_infohashes"`
 }
 
+// Default parser config constants.
+const (
+	defaultMaxNumWant          uint32 = 100
```
jzelinskie: see previous comment

PR #460, opened 2019-10-12
Storage benchmarks

### LINKED ISSUE #447: Update storage benchmarks with a way to clear the storage in between
With redis coming in, we need a way to clear the storage in between benchmarks.

mrd0ll4r:
Fixes #447 
This also adds a small variant of the announce and scrape benchmarks.
From the data we gathered from public trackers, swarms are usually tiny (probably only having one peer), and only a handful of swarms are actually big. So I thought it would be nice to see how the storage handles small and large swarms :)


storage/memory/peer_store_test.go
```diff
@@ -22,30 +22,58 @@ func createNew() s.PeerStore {
...
+func BenchmarkNop(b *testing.B)                         { s.Nop(b, createNew()) }
+func BenchmarkPut(b *testing.B)                         { s.Put(b, createNew()) }
+func BenchmarkPutSpreadPeer(b *testing.B)               { s.PutSpreadPeer(b, createNew()) }
+func BenchmarkPutSpreadInfohash(b *testing.B)           { s.PutSpreadInfohash(b, createNew()) }
+func BenchmarkPutSpreadInfohashSpreadPeer(b *testing.B) { s.PutSpreadInfohashSpreadPeer(b, createNew()) }
+func BenchmarkPutDelete(b *testing.B)                   { s.PutDelete(b, createNew()) }
+func BenchmarkPutDeleteSpreadPeer(b *testing.B)         { s.PutDeleteSpreadPeer(b, createNew()) }
+func BenchmarkPutDeleteSpreadInfohash(b *testing.B)     { s.PutDeleteSpreadInfohash(b, createNew()) }
+func BenchmarkPutDeleteSpreadInfohashSpreadPeer(b *testing.B) {
+	s.PutDeleteSpreadInfohashSpreadPeer(b, createNew())
```
jzelinskie: does gofmt force this into multilines? 😢

storage/redis/peer_store.go
```diff
@@ -810,6 +814,102 @@ func (ps *peerStore) collectGarbage(cutoff time.Time) error {
 	return nil
 }
 
+// Clear clears (almost) all data from redis, given that no concurrent
+// operations are taking place.
+// This uses the same logic as collectGarbage, only that all peers,
+// regardless of age, will be deleted.
+// In the end, only the counter keys remain.
+func (ps *peerStore) Clear() error {
```
jzelinskie: Redis has a concept called databases that are similar to schemas in Postgres or MySQL. We could just create a new database and switch to it. Then we don't have to worry about any keys being in the new context.

mrd0ll4r: I thought about that too! You probably mean we use a database for our stuff, then delete the entire database in `Clear()`, or even just start a new one for the next benchmark without clearing the old one?

storage/memory/peer_store_test.go
mrd0ll4r: Just tried it in the playground: yep. (I'm running goimports, which I'm pretty sure does the same formatting, but I had to test :) )

mrd0ll4r:
Just looked at this again, @jzelinskie can you give me some input on the redis database issue?

PR #461, opened 2019-10-12
travis: add e2e tests with redis storage

### LINKED ISSUE #446: Run e2e tests in CI with different storage implementations
As seen in #433 it would be useful to have e2e tests run on different storage implementations. For redis, it would be good to have it run with an actual, clean redis instance. Travis can do this, we just have to configure it nicely so it uses the right config.

mrd0ll4r:
Fixes #446 

Maybe we should think about putting our build script into an actual script and call that with, like, one parameter that specifies whether or not we want vendored dependencies, because it's getting quite big...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,15 @@ language: go
 go:
   - 1.13.x
 sudo: false
+services:
+  - redis-server
 matrix:
   include:
   # Using vendored dependencies
   - install:
     - go get -u golang.org/x/lint/golint
     - go get -u golang.org/x/tools/cmd/...
+    - go get gopkg.in/mikefarah/yq.v2
     script:
     - go mod download
     - go test -v -race $(go list ./...)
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -2,12 +2,15 @@ language: go
...
 sudo: false
+services:
+  - redis-server
 matrix:
   include:
   # Using vendored dependencies
   - install:
     - go get -u golang.org/x/lint/golint
     - go get -u golang.org/x/tools/cmd/...
+    - go get gopkg.in/mikefarah/yq.v2
```
jzelinskie: shameless selfplug: can we use faq?

```sh
LATEST_FAQ_RELEASE=$(curl -s https://api.github.com/repos/jzelinskie/faq/releases | cat | head -n 10 | grep "tag_name" | cut -d\" -f4)
curl -Lo $GOPATH/bin/faq https://github.com/jzelinskie/faq/releases/download/$LATEST_RELEASE/faq-linux-amd64 && chmod +x $GOPATHbin/faq
```

jzelinskie: Rather, I'd honestly be more fine with checking in benchmark/testing config files in a separate directory and moving our example config into the documentation directory. It'll avoid having brittle scripts in CI doing find/replace.

mrd0ll4r:
Alrighty!
I moved travis stuff into scripts and config files in dist/travis.
Also update example config and redis doc to be in sync and make sense.
Also updated Dockerfile because that still had GO11MODULES=on in it.

mrd0ll4r:
The scripts now have the right permissions in git. No chmod needed anymore :)

PR #462, opened 2019-10-24
Implement HTTP fullscrapes

mrd0ll4r:
Fixes #369 (eventually)

WIP, but need travis

### DIFF
--- a/dist/example_config.yaml
+++ b/dist/example_config.yaml
@@ -68,12 +68,12 @@ chihaya:
     # The maximum number of infohashes that can be scraped in one request.
     max_scrape_infohashes: 50
 
-    # The IP addresses from which a fullscrape will be allowed.
+    # The networks in CIDR notation from which a fullscrape will be allowed.
     # Note: Fullscrapes are potentially expensive and may slow the tracker down.
-    # It is recommended to only allow this from trusted IPs.
+    # It is recommended to only allow this from trusted networks.
     # Whether fullscrapes are supported depends on the storage implementation.
-    fullscrape_whitelisted_ips:
-      - "127.0.0.1"
+    fullscrape_whitelisted_networks:
+      - "127.0.0.1/8"
 
   # This block defines configuration for the tracker's UDP interface.
   # If you do not wish to run this, delete this section.
--- a/frontend/http/parser.go
+++ b/frontend/http/parser.go
@@ -15,12 +15,39 @@ import (
 // If RealIPHeader is not empty string, the value of the first HTTP Header with
 // that name will be used.
 type ParseOptions struct {
-	AllowIPSpoofing          bool     `yaml:"allow_ip_spoofing"`
-	RealIPHeader             string   `yaml:"real_ip_header"`
-	MaxNumWant               uint32   `yaml:"max_numwant"`
-	DefaultNumWant           uint32   `yaml:"default_numwant"`
-	MaxScrapeInfoHashes      uint32   `yaml:"max_scrape_infohashes"`
-	FullScrapeWhitelistedIPs []net.IP `yaml:"fullscrape_whitelisted_ips"`
+	AllowIPSpoofing               bool       `yaml:"allow_ip_spoofing"`
+	RealIPHeader                  string     `yaml:"real_ip_header"`
+	MaxNumWant                    uint32     `yaml:"max_numwant"`
+	DefaultNumWant                uint32     `yaml:"default_numwant"`
+	MaxScrapeInfoHashes           uint32     `yaml:"max_scrape_infohashes"`
+	FullScrapeWhitelistedNetworks []cidrMask `yaml:"fullscrape_whitelisted_networks"`
+}
+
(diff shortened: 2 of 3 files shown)


dist/example_config.yaml
```diff
@@ -68,6 +68,13 @@ chihaya:
     # The maximum number of infohashes that can be scraped in one request.
     max_scrape_infohashes: 50
 
+    # The IP addresses from which a fullscrape will be allowed.
+    # Note: Fullscrapes are potentially expensive and may slow the tracker down.
+    # It is recommended to only allow this from trusted IPs.
+    # Whether fullscrapes are supported depends on the storage implementation.
+    fullscrape_whitelisted_ips:
+      - "127.0.0.1"
```
jzelinskie: Can we make these CIDRs instead of IPs? That'd make this significantly more configurable.

frontend/http/parser.go
```diff
@@ -128,7 +131,31 @@ func ParseScrape(r *http.Request, opts ParseOptions) (*bittorrent.ScrapeRequest,
...
-		return nil, bittorrent.ClientError("no info_hash parameter supplied")
+		// This is a potential fullscrape. Determine the origin IP and check if
+		// it's whitelisted.
+		host, _, err := net.SplitHostPort(r.RemoteAddr)
+		if err != nil {
+			return nil, err
+		}
+		remoteIP := net.ParseIP(host)
+		if remoteIP == nil {
+			return nil, errors.New("unable to parse HTTP remote address as IP")
```
jzelinskie: We parse IPs in a bunch of places. Maybe we should create one error for all of the failure cases across the codebase.

frontend/http/parser.go
```diff
@@ -128,7 +131,31 @@ func ParseScrape(r *http.Request, opts ParseOptions) (*bittorrent.ScrapeRequest,
...
+		}
+
+		found := false
+		for _, whitelistedIP := range opts.FullScrapeWhitelistedIPs {
```
jzelinskie: This code would change if we use CIDR masks, but ideally this is a set lookup and not iteration.

storage/memory/peer_store.go
```diff
@@ -484,26 +489,61 @@ func (ps *peerStore) AnnouncePeers(ih bittorrent.InfoHash, seeder bool, numWant
...
-	shard.RLock()
-
-	swarm, ok := shard.swarms[ih]
-	if !ok {
-		shard.RUnlock()
+	if len(infoHashes) == 0 {
+		// This is a fullscrape.
+		start := time.Now()
+		resp = ps.fullscrapeSwarms(addressFamily)
+		recordFullscrapeDuration(time.Since(start))
```
jzelinskie: Couldn't we pull the metric timing up to the caller of this function? Are we doing that for the non-full-scrape case? The caller can definitely make the check of whether the `len(infoHashes) == 0`

storage/redis/peer_store.go
```diff
@@ -619,42 +619,46 @@ func (ps *peerStore) AnnouncePeers(ih bittorrent.InfoHash, seeder bool, numWant
...
 
-	leechersLen, err := redis.Int64(conn.Do("HLEN", encodedLeecherInfoHash))
-	if err != nil {
-		log.Error("storage: Redis HLEN failure", log.Fields{
-			"Hkey":  encodedLeecherInfoHash,
-			"error": err,
-		})
-		return
-	}
+	for i, ih := range infoHashes {
```
jzelinskie: So redis doesn't support fullscrapes? :(

mrd0ll4r: It's been a while since I wrote this, but my gut feeling is that this had some complexity due to the possibility of asynchronous stuff happening.. I'd probably have to design another `WATCH ... MULTI ... EXEC` block and prove that to be correct. What's the priority on this? How about just a note in the docs that redis currently doesn't do fullscrapes, and an issue to keep track of it?

storage/memory/peer_store.go
mrd0ll4r: Well, we're tracking some request durations in the frontends, if enabled via the config flag.

I kinda see your point, as in: The prometheus metrics for the storage are all storage-internal, as in not visible to a frontend. I just added it here because I thought it would be interesting to see, similar to GC duration.

I expect fullscrapes to be an occasional thing, and I know from some people who didn't enable the request timing in the frontend (because it costs performance).

Opinion?

frontend/http/parser.go
mrd0ll4r: Well, it's kinda messy. Sometimes we return the IP and an error, sometimes we return an IP and a bool... then there's also the whole IP spoofing thing. UDP only copies around bytes/remote socket address and does no checks whatsoever.

I don't think we can unify all of those :sweat_smile:

dist/example_config.yaml
mrd0ll4r: Done

mrd0ll4r:
@jzelinskie I could need some input here too :)

PR #465, opened 2020-01-14
http: allow for customized routes

elotreum:
Update to allow arrays of routes to be passed to the http frontend.
This also supports named parameters as permitted by the
router.

To avoid external dependencies in the middleware, a RouteParam and
RouteParams type was added to the bittorrent package.

Note: this eliminates the need for "enable_legacy_php_urls", as
the the additional route could be added to the route array. However,
this may be considered a breaking change.

### DIFF
--- a/bittorrent/params.go
+++ b/bittorrent/params.go
@@ -54,6 +54,22 @@ type QueryParams struct {
 	infoHashes []InfoHash
 }
 
+type routeParamsKey struct{}
+
+// RouteParamsKey is a key for the context of a request that
+// contains the named parameters from the http router
+var RouteParamsKey = routeParamsKey{}
+
+// RouteParam is a type that contains the values from the named parameters
+// on the route
+type RouteParam struct {
+	Key   string
+	Value string
+}
+
+// RouteParams is a collection of RouteParam instances
+type RouteParams []RouteParam
+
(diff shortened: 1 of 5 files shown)


bittorrent/params.go
```diff
@@ -54,6 +54,22 @@ type QueryParams struct {
 	infoHashes []InfoHash
 }
 
+type routeParamsKey struct{}
+
+// RouteParamsKey is a key for the context of a request that
+// contains the named parameters from the http router
```
mrd0ll4r: Godoc comments should be full sentences, including a period at the end.

mrd0ll4r:
Oh, and you need to update the config files in `dist/travis/` to get the e2e tests to work again :)

bittorrent/params.go
elotreum: Thanks for the catch. Latest push fixes this.

elotreum:
> In the long term, if you want, we could shave off a ton of allocations by recycling a custom context-ish type for requests, but that's not for this PR.

Not sure I'm following here. Where are the allocations happening and are you thinking that we could just implement a `map[string]interface{}` to pass along on context from the request, rather than the `context.WithValue` call?

elotreum:
> Oh, and you need to update the config files in `dist/travis/` to get the e2e tests to work again :)

Done.

PR #466, opened 2020-01-20
add helper method to retrieve RouteParam by name

elotreum:
Add a helper method to the RouteParams to retrieve a value by name

### DIFF
--- a/bittorrent/params.go
+++ b/bittorrent/params.go
@@ -72,6 +72,10 @@ type RouteParams []RouteParam
 
 // ByName returns the value of the first RouteParam that matches the given
 // name. If no matching RouteParam is found, an empty string is returned.
+// In the event that a "catch-all" parameter is provided on the route and
+// no value is matched, an empty string is returned. For example: a route of
+// "/announce/*param" matches on "/announce/". However, ByName("param") will
+// return an empty string.
 func (rp RouteParams) ByName(name string) string {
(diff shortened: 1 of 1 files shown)


bittorrent/params.go
```diff
@@ -70,6 +70,17 @@ type RouteParam struct {
 // RouteParams is a collection of RouteParam instances.
 type RouteParams []RouteParam
 
+// ByName returns the value of the first RouteParam that matches the given
+// name. If no matching RouteParam is found, an empty string is returned.
```
mrd0ll4r: From the [httprouter README](https://github.com/julienschmidt/httprouter#catch-all-parameters) I see a possible confusion in this scenario: If we use a catch-all parameter like so:
```
/announce/*param
```
and a client requests `/announce/`, then `httprouter` will match the above route and set `param` to `""` (the empty string). Is that correct?
If yes, a consumer of `ByName` can not distinguish between "param not found" and "param set to empty string".

That being said, I think this is expected behaviour (for catch-all parameters). Please add a godoc comment that explains the situation :)

jzelinskie:
Is this being called anywhere or is it just for consumers of the bittorrent package in other codebases?

elotreum:
@jzelinskie This will be for consumers of the bittorrent package, specifically by those consumers that implement their own middleware that rely on a route configured with named parameters.

PR #468, opened 2020-02-15
Replace TravisCI with GitHub Workflows

jzelinskie:


.github/workflows/go.yaml
```diff
@@ -0,0 +1,132 @@
...
+      run: |
+        # temporary fix for https://github.com/actions/setup-go/issues/14
+        echo "##[set-env name=GOPATH;]$(dirname $GITHUB_WORKSPACE)"
+        echo "##[add-path]$(dirname $GITHUB_WORKSPACE)/bin"
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Build
+      run: go build -v ./cmd/...
```
mrd0ll4r: Does this use modules by default now?

mrd0ll4r:
Can I see this in action somewhere? :)

.github/workflows/go.yaml
jzelinskie: Yep

jzelinskie:
https://github.com/jzelinskie/chihaya/pull/2

PR #472, opened 2020-03-13
storage/redis: reimplement using scripts

### LINKED ISSUE #471: infohashes_count goes into negative values
Hello,

I switched from a 20k infohashes_count memory storage single tracker
to two trackers connecting to one redis (residing on tracker 1, tracker 2 has redis slave for backup reasons, since multi master with dynomite didn't work), both have the same private_key.

Everything is default, except for:
  announce_interval: 5m
  min_announce_interval: 2m
  storage:
     name: redis
     config:
       peer_lifetime: 6m
       redis_broker: "redis://authpw@redisaddress:6379/0"

Current stats show:
chihaya_storage_infohashes_count -9149
chihaya_storage_leechers_count 19898
chihaya_storage_seeders_count 10761

Is the interval way too low, so there are weird glitches causing this behavour?
I kept the min announce to 5m beforehand, but that caused issues of announcing not working sometimes.

Best regards,
SolSoCoG

mrd0ll4r:
This possibly fixes #471, and also possibly increases performance, but who knows...


storage/redis/peer_store.go
```diff
@@ -659,51 +649,62 @@ func (ps *peerStore) ScrapeSwarm(ih bittorrent.InfoHash, af bittorrent.AddressFa
...
+		if err != nil {
+			log.Error("storage: HGETALL failed", log.Fields{
+				"key":   swarmKey,
+				"error": err,
+			})
+			return err
+		}
+
+		var peerKey serializedPeer
+		//var removedPeerCount int64
```
jzelinskie: forgot this?

storage/redis/peer_store.go
```diff
@@ -659,51 +649,62 @@ func (ps *peerStore) ScrapeSwarm(ih bittorrent.InfoHash, af bittorrent.AddressFa
...
+						log.Error("storage: gcPeer script execution failed", log.Fields{
+							"error": err,
+						})
+						return err
+					}
+
+				}
+			} else { // key
+				peerKey = serializedPeer([]byte(field))
+			}
```
jzelinskie: I think you should put this as the first case, as it is small. Then make the `// key` and `// value` comments full sentences.

storage/redis/scripts.go
```diff
@@ -0,0 +1,238 @@
...
+		swarmKey = seederSwarmKey(peer.IP.AddressFamily, infoHash)
+		countKey = seederCountKey(peer.IP.AddressFamily)
+	} else {
+		swarmKey = leecherSwarmKey(peer.IP.AddressFamily, infoHash)
+		countKey = leecherCountKey(peer.IP.AddressFamily)
+	}
+	swarmCountsKey := swarmCountsKey(peer.IP.AddressFamily)
+
+	infoHashString := infoHash.String()
+	peerKey := newPeerKey(peer)
```
jzelinskie: just inline these in the return statement

storage/redis/peer_store.go
```diff
@@ -1,30 +1,36 @@
...
 //
-// - IPv{4,6}_{L,S}_infohash
-//	To save peers that hold the infohash, used for fast searching,
-//  deleting, and timeout handling
+// In general, each swarm is tracked in two hashes per address family:
 //
-// - IPv{4,6}
-//  To save all the infohashes, used for garbage collection,
-//	metrics aggregation and leecher graduation
+// - swarm_<infohash>_IPv{4|6}_{seeders|leechers} -> <peer key> -> <last modified time>
```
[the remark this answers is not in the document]
mrd0ll4r: I guess.. I renamed them to be more verbose because it took me so long to figure out what was happening with the short names :)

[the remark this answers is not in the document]
mrd0ll4r: Ah, I found something better: Key names don't need to be ASCII or whatever, so I can use the 20-byte infohash instead of the 40-byte hex representation...

storage/redis/peer_store.go
```diff
@@ -263,32 +279,72 @@ type peerStore struct {
...
+	// we can exploit that "seeders" is shorter than "leechers"
+	seeder := len(swarmKey) == 59
+
+	var af bittorrent.AddressFamily
+	if addressFamilyPart == "IPv4" {
+		af = bittorrent.IPv4
+	} else if addressFamilyPart == "IPv6" {
+		af = bittorrent.IPv6
+	} else {
+		panic(fmt.Sprintf("unknown address family %q", addressFamilyPart))
```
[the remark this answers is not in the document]
mrd0ll4r: I opted for `%#v` and convert it to a byte slice now, because it could be binary

storage/redis/peer_store.go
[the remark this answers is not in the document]
mrd0ll4r: I used the same trick with the infohash peer count keys. Btw, in `redis-cli` they are displayed like this now: `"swarm_\xb2\xbcP\\\xe9\xca\x92\r\x1a\xe6\xb7\x0e\xc5\xf8\xea A\x1ab\xbd_IPv4_leechers"`, so that works...

mrd0ll4r:
PTAL

storage/redis/peer_store.go
```diff
@@ -263,32 +279,145 @@ type peerStore struct {
...
+	return []bittorrent.AddressFamily{bittorrent.IPv4, bittorrent.IPv6}
 }
 
-func (ps *peerStore) leecherInfohashKey(af, ih string) string {
-	return af + "_L_" + ih
+// Constants used for constructing/deconstructing swarm keys.
+const (
+	swarmPrefix                = "swarm_"
+	seedersSuffix              = "_seeders"
+	leechersSuffix             = "_leechers"
```
jzelinskie: Can we go back to using the smallest constants possible here? Keys like this are readable, but are going to eat memory.

storage/redis/peer_store.go
```diff
@@ -717,89 +792,39 @@ func (ps *peerStore) collectGarbage(cutoff time.Time) error {
...
-		// list all infohashes in the group
-		infohashesList, err := redis.Strings(conn.Do("HKEYS", group))
+	var (
+		cursor int64
+		items  []string
+	)
+
+	// SCAN through all of our swarm keys.
+	for {
+		values, err := redis.Values(conn.Do("SCAN", cursor, "MATCH", "swarm_*", "COUNT", 50))
```
jzelinskie: This is okay, as long as there's one instance of Chihaya. Once there are multiple trying to GC the same redis instance/cluster, this strategy is non-ideal. I think calculating a shard size based on the total size of the database, and then choosing one at random then GCing all of those keys might be better.

storage/redis/peer_store.go
```diff
@@ -263,32 +279,145 @@ type peerStore struct {
...
+	leecherSwarmKeyLength      = len(swarmPrefix) + 20 + 1 + addressFamilyLength + len(leechersSuffix)
+	seederSwarmKeyLength       = len(swarmPrefix) + 20 + 1 + addressFamilyLength + len(seedersSuffix)
+	swarmKeyInfohashStart      = len(swarmPrefix)
+	swarmKeyInfohashEnd        = swarmKeyInfohashStart + 20
+	swarmKeyAddressFamilyStart = swarmKeyInfohashEnd + 1
+	swarmKeyAddressFamilyEnd   = swarmKeyAddressFamilyStart + addressFamilyLength
+)
+
+// The length of IPv4/IPv6.
+const addressFamilyLength = len(bittorrent.IPv4String)
```
jzelinskie: why not just use the byte/uint8? that seems like it'll save memory and a bunch of branches

storage/redis/peer_store.go
jzelinskie: On another note -- is there a reason we don't do this part in the script as well to avoid round trips?

mrd0ll4r: SCAN is nondeterministic, and Redis doesn't allow scripts to have write commands (such as deleting stuff) after anything nondeterministic, in order not to break replication.

I mostly get what you're saying in your first comment, i.e. instead of processing everything with a `swarm_` prefix, better pick one of `swarm_<single byte>` at random and GC that. (The first byte is the largest shard we can get.)
I don't like it, because:

1.  As the largest shards are 1/256th of the whole keyspace, we would need to GC 256 times as often, on average, to get the same rate of garbage collection as before. (or do multiple shards in one GC)
2. This rate changes with the number of chihaya instances running.
3. We could try to have  a counter, or volatile keys, or some other mechanism of counting the number of active chihaya instances, but this is non-trivial. (This is an interesting problem and I kind of want to work on it, but not for this PR.)
4. The added randomness is good for distributed systems like this, but I would argue we don't need it: The SCAN command is already nondeterministic, i.e. it depends on some internal state of the redis instance. There are [a bunch of properties](https://redis.io/commands/scan#scan-guarantees) around what keys are returned by SCAN, which make me think that this is generally a very elegant solution to our problem.

storage/redis/peer_store.go
mrd0ll4r: I have now boiled this down to the shortest version possible, I believe.

storage/redis/peer_store.go
jzelinskie: >The added randomness is good for distributed systems like this, but I would argue we don't need it: The SCAN command is already nondeterministic, i.e. it depends on some internal state of the redis instance. There are a bunch of properties around what keys are returned by SCAN, which make me think that this is generally a very elegant solution to our problem.

SCAN is nondeterministic because it's implemented as a cursor in a database that isn't MVCC. Thus parallel operations will change values out from underneath it while you're iterating. That's all it means; it still iterates in the same _order_ across clients. I've both [read the code](https://github.com/antirez/redis/blob/a20c206de7cb3ccd176b05b883e11b03c34db417/src/db.c#L886) and tested this locally to confirm that this is the case. The guarantees listed in that doc are the exact same in our iteration of the keyspace for the memory implementation.

>This rate changes with the number of chihaya instances running.

This is already the case -- if I run two instances of chihaya, the garbage collection on each instance is set to run at the same interval, but different offsets based on when the instance was started. If they were both started around the same time, they're basically going to iterate over the whole database one right after the other which would make the second instance functionally useless while also being expensive on the database.

In order to maximize the usefulness of each instance's garbage collection, the goal should be to have a cluster-wide solution to garbage collection. A solution could roughly look like one of two options: elect only one garbage collection process to run across the deployment or devise a strategy for collaboration across instances. If we're clever that collaboration need not require coordination, which is complicated to implement.

In an [unrelated project](https://github.com/quay/quay/blob/531b97c4f90e11055eaa8a6621219623b603e293/util/migrate/allocator.py), we have a means of coordination-free work for backfilling database values. You can keep spinning up more workers and the progress will get faster and faster. For this implementation, the requirements are idempotent work and a uniform keyspace indexed by integers that you can use as bounds.

While this is not directly applicable, I believe that thinking about a strategy along these lines could provide a more ideal solution. Thus, I'd rather discuss and explore this train of thought before compromising and doing something like staggering when GC starts or using a global lock to elect the next person to GC.

jzelinskie:
Hey, I don't think we should leave this blocked on the GC design.
We can always fix that in a follow up to be more efficient.

PR #476, opened 2020-04-05
updating go deps

moistari:
Updates to latest go dependencies, fixing a minor API change for
prometheus.

Should have significant impact on performance.


jzelinskie:
Looks good, can we also bump to Go 1.14 while we're at it?

moistari:
Sorry, I didn't see your comment here until just now. Sure, can bump the Go version, but it wouldn't really affect anything? The purpose of the `go <version>` data in the `go.mod` is to set the minimum Go version required, which is used (for example) when the target code base uses a language feature or API change in the standard library that was introduced after the pinned version.

But happy to oblige as I'm no supporter for backwards compatibility in build tools. I've pushed the requested changes.

jzelinskie:
Cool, this looks good. I don't really like checking in gitignore files, because they end up a dumping ground for whatever anyone fancies to ignore. Thoughts @mrd0ll4r ?

mrd0ll4r:
Yup, agree with you @jzelinskie

jzelinskie:
I've automated this work in the future with dependabot. If you could just remove the gitignore file from your PR, I'll merge this and close all of the other automated PRs.

jzelinskie:
Thanks for this PR. We ended up merging #489 because I accidentally added conflicts to this PR when experimenting with dependabot.

PR #483, opened 2020-04-29
.github/workflows: only run dependabot CI on PR

jzelinskie:
Because dependabot creates branches on the origin repository, it triggers a CI run for both "push" and "pull request" unless this patch is applied.


mrd0ll4r:
...it runs recursively now?

PR #492, opened 2020-05-07
pkg/log: add type field for errors

jzelinskie:
This should help a lot for debugging new errors.


mrd0ll4r:
Nice, yes, that seems like a very good idea!
For the record: Even though `%T` [uses reflection](https://github.com/golang/go/blob/master/src/fmt/print.go#L656), the call to `LogFields` should only be done for errors we actually log, i.e. this should not add cost for disabled debug logging or anything like that.

PR #515, opened 2021-02-27
Move profiling into the metrics server

jzelinskie:


mrd0ll4r:
LGTM, but we have another config file in the repository, in dist/helm/chihaya/values.yaml, should we/do we keep that up to date as well?

jzelinskie:
Fixed up the Helm and even added CI for it.

mrd0ll4r (approved):
Oboi, lots of devops stuff I don't fully understand... But I do see the test action working, looks good to me :)

PR #540, opened 2021-07-05
*: migrate to netaddr

jzelinskie:
I'm sorry for the huge commit. This ended up touching far more than I thought it would.
I'm interested in seeing some perf numbers on real workloads before merging.

This change impacted almost the entire codebase:

- All uses of net.IP are now netaddr.IP or netaddr.IPPort
- Scrape now passes an entire peer
- Testing in bittorrent package refactored
- Parsing and API changes to Peers, PeerIDs, ClientIDs
- Redis Storage now uses shorter keys


storage/storage.go
```diff
@@ -102,7 +104,7 @@ type PeerStore interface {
 	// filling the Snatches field is optional.
 	//
 	// If the Swarm does not exist, an empty Scrape and no error is returned.
-	ScrapeSwarm(infoHash bittorrent.InfoHash, addressFamily bittorrent.AddressFamily) bittorrent.Scrape
+	ScrapeSwarm(infoHash bittorrent.InfoHash, ip netaddr.IP) bittorrent.Scrape
```
mrd0ll4r, line 107: I wonder if this might impact performance because `netaddr.IP` is 3 words wide on a 64bit system (I think?) whereas `AddressFamily` should only be one word. I guess we should do some benchmarking at some point...

storage/redis/peer_store.go
```diff
@@ -267,28 +238,77 @@ type peerStore struct {
...
-}
+	if seed {
+		b.WriteString("S")
+	} else {
+		b.WriteString("L")
+	}
 
-func (ps *peerStore) infohashCountKey(af string) string {
-	return af + "_infohash_count"
+	b.WriteString(infoHash.String())
```
mrd0ll4r, line 264: We could use `RawString` to save 20 bytes, which is nice, but makes the keys non-human readable. I think that's what we used before, semantically?

storage/redis/peer_store.go
```diff
@@ -267,28 +238,77 @@ type peerStore struct {
...
 
-func (ps *peerStore) leecherInfohashKey(af, ih string) string {
-	return af + "_L_" + ih
-}
+// Key that stores the peers for an InfoHash.
+//
+// "4S0102030405060708090a0b0c0d0e0f1011121314" => IPv4 Seeder with Peer ID 0102030405060708090a0b0c0d0e0f1011121314
+// "6L0204090405060708060a0b0c0d0e0f1011121328" => IPv6 Seeder with Peer ID 0204090405060708060a0b0c0d0e0f1011121328
+func peerKey(infoHash bittorrent.InfoHash, seed bool, ip netaddr.IP) string {
+	var b strings.Builder
```
mrd0ll4r, line 248: Maybe `Grow` to the appropriate size :)

storage/redis/peer_store.go
```diff
@@ -267,28 +238,77 @@ type peerStore struct {
...
+	if seed {
+		b.WriteString("S")
+	} else {
+		b.WriteString("L")
+	}
+
+	return b.String()
 }
 
-func (ps *peerStore) leecherCountKey(af string) string {
```
mrd0ll4r, line 290: There's a doc in `docs/storage/redis.md` that describes these keys and whatnot, that needs to be updated as well.

storage/redis/peer_store.go
```diff
@@ -267,28 +238,77 @@ type peerStore struct {
...
-	return af + "_L_count"
+// Key that stores the total number of infohashes that has peers of an IP
+// version.
+//
+// "I4" => number of IPv4 InfoHashes
+// "I6" => number of IPv6 InfoHashes
+func infohashCountKey(ip netaddr.IP) string {
+	var b strings.Builder
```
mrd0ll4r, line 299: I wonder if the `Builder` is necessary or actually adds overhead. We know this is gonna be two bytes, so we might as well just allocate a slice for that.

storage/redis/peer_store.go
```diff
@@ -267,28 +238,77 @@ type peerStore struct {
...
+var addressFamilies = [2]string{"4", "6"}
 
-func (ps *peerStore) leecherInfohashKey(af, ih string) string {
-	return af + "_L_" + ih
-}
+// Key that stores the peers for an InfoHash.
+//
+// "4S0102030405060708090a0b0c0d0e0f1011121314" => IPv4 Seeder with Peer ID 0102030405060708090a0b0c0d0e0f1011121314
+// "6L0204090405060708060a0b0c0d0e0f1011121328" => IPv6 Seeder with Peer ID 0204090405060708060a0b0c0d0e0f1011121328
+func peerKey(infoHash bittorrent.InfoHash, seed bool, ip netaddr.IP) string {
```
mrd0ll4r, line 247: I think we should rename this, to `swarmKey` maybe.

storage/redis/peer_store.go
```diff
@@ -348,32 +372,33 @@ func (ps *peerStore) PutSeeder(ih bittorrent.InfoHash, p bittorrent.Peer) error
...
+	addressFamily := pk[0]
 	ct := ps.getClock()
 
 	conn := ps.rb.open()
 	defer conn.Close()
 
 	conn.Send("MULTI")
-	conn.Send("HSET", encodedSeederInfoHash, pk, ct)
-	conn.Send("HSET", addressFamily, encodedSeederInfoHash, ct)
+	conn.Send("HSET", pk, p.String(), ct)
```
mrd0ll4r, line 384: This wastes some memory, same as with the memory storage.

storage/redis/peer_store.go
```diff
@@ -348,32 +372,33 @@ func (ps *peerStore) PutSeeder(ih bittorrent.InfoHash, p bittorrent.Peer) error
...
+	conn.Send("HSET", addressFamily, pk, ct)
 	reply, err := redis.Int64s(conn.Do("EXEC"))
 	if err != nil {
 		return err
 	}
 
 	// pk is a new field.
```
mrd0ll4r, line 391: These comments talk about wrong fields now

storage/redis/peer_store.go
```diff
@@ -743,23 +762,23 @@ func (ps *peerStore) collectGarbage(cutoff time.Time) error {
 					}
 					if mtime <= cutoffUnix {
 						log.Debug("storage: deleting peer", log.Fields{
-							"Peer": decodePeerKey(pk).String(),
+							"Peer": bittorrent.MustParsePeer(peerStr),
```
mrd0ll4r, line 765: I know it's been like this before, but I dislike that we do the parsing even though debug logging might be disabled...

mrd0ll4r (changes requested):
Some performance and memory things I disagree with, and a lot of small things :)

storage/storage.go
jzelinskie: I don't think we'll see any appreciable difference because it's all on the stack.

storage/redis/peer_store.go
jzelinskie: Yep, I actually swapped these around and forgot to change this back. It's not supposed to be hex here.

storage/redis/peer_store.go
jzelinskie: I'd like to switch us to zerolog, which is what I use at work. It's more ergonomic than logrus and would let us drop our own log package because the zerolog API is almost exactly the same as what we've built.

bittorrent/bittorrent.go
```diff
@@ -187,65 +232,49 @@ type Scrape struct {
...
+// MustParsePeer parses a Peer in the format `<PeerID>@<IP>:<Port>` and panics
+// if it fails.
+func MustParsePeer(s string) Peer {
+	return Peer{
+		ID:     PeerIDFromRawString(s[:20]),
+		IPPort: netaddr.MustParseIPPort(s[21:]),
+	}
+}
+
+// String implements fmt.Stringer for a readable representation of a Peer.
```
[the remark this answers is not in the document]
jzelinskie: I actually started doing this, but it became too much at once, so I switched back.

frontend/udp/parser.go
```diff
@@ -91,15 +96,18 @@ func ParseAnnounce(r Request, v6Action bool, opts ParseOptions) (*bittorrent.Ann
...
 
-	ip := r.IP
+	ip := r.IPPort.IP()
 	ipProvided := false
 	ipbytes := r.Packet[84:ipEnd]
 	if opts.AllowIPSpoofing {
-		// Make sure the bytes are copied to a new slice.
-		copy(ip, net.IP(ipbytes))
+		var ok bool
+		ip, ok = netaddr.FromStdIP(net.IP(ipbytes))
```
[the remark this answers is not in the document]
jzelinskie: Because net.IP is just a typedef of []byte, I'm pretty sure the conversion here is a no-op.

storage/redis/peer_store.go
```diff
@@ -267,28 +239,74 @@ type peerStore struct {
...
-}
+// Key that stores the peers for an InfoHash.
+//
+// "4S0102030405060708090a0b0c0d0e0f1011121314" => IPv4 Seeder with Peer ID 0102030405060708090a0b0c0d0e0f1011121314
+// "6L0204090405060708060a0b0c0d0e0f1011121328" => IPv6 Seeder with Peer ID 0204090405060708060a0b0c0d0e0f1011121328
+func swarmKey(infoHash bittorrent.InfoHash, seed bool, ip netaddr.IP) string {
+	var b strings.Builder
+	switch {
+	case ip.Is4(), ip.Is4in6():
+		b.Grow(1 + 1 + 4) // "4" + "S"/"L" + IPv4
```
mrd0ll4r, line 252: Missing 20 bytes for the actual infohash 🙃

mrd0ll4r (changes requested):
I just saw the one buffer growth thing. Apart from that: It's a big one, not sure if I saw everything... maybe we can/should run the various benchmarking things? `tbomb` for just raw performance, and the storage memory usage thing (although I haven't tried that out in ages and don't know if it still works).

storage/redis/peer_store.go
jzelinskie: nice catch

I also noticed that in PutLeecher it does not check if the infohash was new. It's apparently missing a 

```go
	if reply[1] == 1 { // The Infohash Key (and thus the infohash) was new.
		_, err = conn.Do("INCR", infohashCountKey(ip))
		if err != nil {
			return err
		}
	}
```

Is that on purpose? Should I add this?

mrd0ll4r: I believe that's on purpose, because for redis we only count infohashes with at least one seeder. I don't remember why I did it like that, but I assume there was some reason for it...maybe it would have been difficult to correctly determine whether a seeder-only or leecher-only swarm already exists or something, not sure.

Aranjedeath:
Yay!!! Thank you all. @mrd0ll4r we are running this branch in production currently :) I use regular memory storage backend, so redis issue does not affect me.

mrd0ll4r:
Oh nice @Aranjedeath ! I can see the reboot and most things look fine so far. One thing I can't see is CPU usage (and request durations iirc), let us know how that works out :)

Aranjedeath:
> Oh nice @Aranjedeath ! I can see the reboot and most things look fine so far. One thing I can't see is CPU usage (and request durations iirc), let us know how that works out :)

IPV6 has request timing enabled, we disabled it on v4 previously for performance :)

Aranjedeath:
> Oh nice @Aranjedeath ! I can see the reboot and most things look fine so far. One thing I can't see is CPU usage (and request durations iirc), let us know how that works out :)

OK I think we have performance details in v4 now :)

jzelinskie:
I pushed a commit that migrates us to zerolog in this branch. It's for convenience for testing. If you want me to open it as another PR after this, I can.

bittorrent/bittorrent.go
```diff
@@ -14,6 +14,7 @@ import (
 	"inet.af/netaddr"
 
 	"github.com/chihaya/chihaya/pkg/log"
+	"github.com/jzelinskie/stringz"
```
mrd0ll4r, line 14: 👀

bittorrent/sanitize.go
```diff
@@ -48,8 +50,6 @@ func SanitizeScrape(r *ScrapeRequest, maxScrapeInfoHashes uint32) error {
 		r.InfoHashes = r.InfoHashes[:maxScrapeInfoHashes]
 	}
 
-	log.Debug("sanitized scrape", r, log.Fields{
-		"maxScrapeInfoHashes": maxScrapeInfoHashes,
-	})
+	log.Debug().Uint32("maxScrapeInfoHashes", maxScrapeInfoHashes).Msg("sanitized scrape")
```
mrd0ll4r, line 53: This is missing the sanitized request

bittorrent/sanitize.go
```diff
@@ -25,19 +25,21 @@ func SanitizeAnnounce(r *AnnounceRequest, maxNumWant, defaultNumWant uint32) err
...
-	log.Debug("sanitized announce", r, log.Fields{
-		"maxNumWant":     maxNumWant,
-		"defaultNumWant": defaultNumWant,
-	})
+	// If it's an IPv4 address in in IPv6 form, convert to IPv4, for ubiquity.
+	if ip.Is4in6() {
+		r.IPPort = r.IPPort.WithIP(netaddr.IPFrom4(ip.As4()))
+	}
+
+	log.Debug().
```
mrd0ll4r, line 39: This is also missing the request

storage/storage_bench.go
```diff
@@ -435,7 +437,7 @@ func AnnounceSeeder1kInfohash(b *testing.B, ps PeerStore) {
 // ScrapeSwarm can run in parallel.
 func ScrapeSwarm(b *testing.B, ps PeerStore) {
 	runBenchmark(b, ps, true, putPeers, func(i int, ps PeerStore, bd *benchData) error {
-		ps.ScrapeSwarm(bd.infohashes[0], bittorrent.IPv4)
+		ps.ScrapeSwarm(bd.infohashes[0], bd.peers[0].IPPort.IP())
```
mrd0ll4r, line 440: I think these should be pulled outside the runner function

storage/storage_bench.go
```diff
@@ -445,7 +447,7 @@ func ScrapeSwarm(b *testing.B, ps PeerStore) {
 // ScrapeSwarm1kInfohash can run in parallel.
 func ScrapeSwarm1kInfohash(b *testing.B, ps PeerStore) {
 	runBenchmark(b, ps, true, putPeers, func(i int, ps PeerStore, bd *benchData) error {
-		ps.ScrapeSwarm(bd.infohashes[i%1000], bittorrent.IPv4)
+		ps.ScrapeSwarm(bd.infohashes[i%1000], bd.peers[0].IPPort.IP())
```
mrd0ll4r, line 450: same here

mrd0ll4r (changes requested):
Every time I look at this I find some new small things..
I think it's good overall and I want it merged, but the PR is too big for me to confidently review it 🙃 

Can we somehow split it up maybe?

jzelinskie:
I'll break this down and rework everything now that netaddr got added to the standard library as `net/netip`

PR #549, opened 2022-01-02
.github: split workflows into build/lint, add new linters

jzelinskie:
This fixes a `go vet` failure for using an unbuffered channel with
os.Signal. This also simplifies platform-specific code.

GoFmt added new forward-compatible build tags.


mrd0ll4r (approved):
LGTM, apart from the one check not passing

jzelinskie:
@mrd0ll4r PTAL
I added a bunch more linters and fixed issues with them.

Notably, it caught the fact that we were allocating despite using sync.Pool.

frontend/udp/bytepool/bytepool.go
```diff
@@ -11,24 +11,27 @@ type BytePool struct {
 func New(length int) *BytePool {
 	var bp BytePool
 	bp.Pool.New = func() interface{} {
-		return make([]byte, length)
+		b := make([]byte, length)
+		return &b
```
mrd0ll4r, line 15: Huh. I went down the rabbit hole, which also lead me [here](https://github.com/mailru/easyjson/issues/157).

If I understand this correctly, the comment on staticcheck.io is a bit misleading. While it's true that "passing a value that is not a pointer to a function that accepts an interface, the value needs to be placed on the heap", I did not immediately see how this affects us, as all the network functions we use take byte slices, not interfaces. From above github discussion, however, I'm now under the impression that `(*sync.Pool) Put(x interface{})` itself is the culprit...

Very weird. But nice catch!

mrd0ll4r (changes requested):
Just one minor thing.

bittorrent/params.go
```diff
@@ -189,7 +189,7 @@ func (qp *QueryParams) String(key string) (string, bool) {
 
 // Uint64 returns a uint parsed from a query. After being called, it is safe to
```
mrd0ll4r, line 190: Godoc doesn't fit, and apparently none of the linters caught it 🙃 My frustration with Go linters is growing

mrd0ll4r (changes requested):

bittorrent/params.go
jzelinskie: Yeah, for some reason the golangci-lint authors believe that no one follows these conventions so they implicitly disable these lint rules.

mrd0ll4r (approved):

PR #557, opened 2022-02-14
Add database storage support (postgres and sqlite).

nitrix:
Hello,

This PR adds a new database storage and also registers two drivers: `postgres` and `sqlite`.

The implementation is done using `gorm`, an ORM for Go, which in theory allows us to add easily more SQL-based engines (`mysql`, `clickhouse`, etc). There are 4 tables (`ipv4_leechers`, `ipv4_seeders`, `ipv6_leechers` and `ipv6_seeders`. It re-uses the concept of a serialized peerKey like the other storages do.

A consequence of the current design makes finding all the active unique infohashes fairly slow, so it's strongly recommended to use a larger value for `prometheus_reporting_interval` until that gets revisited.

I'm personally running a cockroachdb cluster (postgres compatible) and my configuration looks like this:

```
   storage:
    name: "postgres"
    config:
      gc_interval: "3m"
      peer_lifetime: "31m"
      prometheus_reporting_interval: "1m"
      dsn: postgresql://user@host:port/db?sslmode=verify-full&sslrootcert=ca.crt&sslcert=user.crt&sslkey=user.key
```

Maybe that's useful to someone.

Cheers,
nitrix.


dist/example_config.yaml
```diff
@@ -164,6 +163,34 @@ chihaya:
...
+
+  #     # The amount of time until a peer is considered stale.
+  #     # To avoid churn, keep this slightly larger than `announce_interval`
+  #     peer_lifetime: "31m"
+
+  #     # The interval at which metrics about the number of infohashes and peers
+  #     # are collected and posted to Prometheus. Disabled by default.
+  #     prometheus_reporting_interval: "1s"
+
+  #     # In lightweight mode, the announces and scrapes pretends there are no peers in the swarm.
```
mrd0ll4r, line 184: From the implementation, this seems to only apply to scrapes (which I think is correct). If that is indeed intended, maybe rename this to `ignore_scrapes` or something like that?

mrd0ll4r (changes requested):
Very interesting stuff! I took a first look, without running it, and noted down some comments.

How is this performance-wise? How large (roughly) is your setup?

nitrix:
@mrd0ll4r I was very optimistic when I did this.

**Multiple issues have crop up since.**

You're seeing in the PR my sad attempts at improving things with the `lightweight` idea and changes to scrapes, but I think it's a little doomed.

The scrapes are what's killing the entire system in practice (I didn't know they happen also on announces). There's a lot of concurrent `SELECT COUNT(*)`s while there are `INSERT`s happening on the same tables... it makes the SQL server want to commit seppuku.

I think there'll need to be a dedicated table with counts, though it's tricky to do it atomically without generating a ton of transactions.

It took my Kubernetes cluster down after increasing the canary to just 2% of the normal load :(

The idea is there but it'll require some refining before it can scale.

Cheers :)

nitrix:
This is great feedback.
I'm going to close the PR because it's becoming a mixed bag of too many different things.
Will re-open another one when I have something that looks less like a draft.

nitrix:
Oh, also, if I can't get this to perform nicely, I'll pursue an in memory-based approach that parodically sync peers between the nodes.

mrd0ll4r:
Don't worry too much about performance -- I was just interested. It's unsurprising that an in-memory solution would be faster, but using a database could have other upsides, like high availability.

jzelinskie:
Feel free to keep the PR open but set it as a draft. That way you can still run changes against CI and there will be some documentation on your progress.

Have you tried the redis storage implementation? It might be more fruitful to spend some time optimizing it and having all of your tracker instances sharing a redis (or redis cluster).

PR #558, opened 2022-02-18
Proof-of-concept clustering

nitrix:
## What

This PR adds the ability to link together multiple chihaya nodes to form a cluster, splitting the workload of servicing the requests amongst themselves.

## How
 
Each node runs its own swarm using [memory storage](https://github.com/chihaya/chihaya/tree/main/storage/memory) and the incoming requests are routed to specific nodes in a stable manner based on the torrent's infohash and the amount of nodes.

## High-availability

There isn't a mechanism to ensure that the data is available on other nodes. If a node leaves the cluster, its data goes away with it. Considering that peers have to re-announce themselves periodically, this is only a minor nuisance sometimes waiting longer to obtain peers.

## Scaling

Adding or removing nodes to the cluster directly affects the partitioning of the infohashes. The addition of 1 node to a 5 nodes cluster would cause a 20% disruption to all nodes, where some of the requests originally meant for them are now being serviced by a different node, which because it just joined, wont have much data to provide for a little while.

For all these reasons, it's preferable if the cluster doesn't scale too frequently.

## Configuration

The configuration provides the same fields that the memory storage does, plus new ones.

```
  # This block defines configuration used for cluster storage.
  # storage:
  #   name: cluster
  #   config:
  #     # The frequency which stale peers are removed.
  #     # This balances between
  #     # - collecting garbage more often, potentially using more CPU time, but potentially using less memory (lower value)
  #     # - collecting garbage less frequently, saving CPU time, but keeping old peers long, thus using more memory (higher value).
  #     gc_interval: "3m"

  #     # The amount of time until a peer is considered stale.
  #     # To avoid churn, keep this slightly larger than `announce_interval`
  #     peer_lifetime: "31m"

  #     # The number of partitions data will be divided into in order to provide a
  #     # higher degree of parallelism.
  #     shard_count: 1024

  #     # The interval at which metrics about the number of infohashes and peers
  #     # are collected and posted to Prometheus.
  #     prometheus_reporting_interval: "1s"

  #     # The address and port the cluster will be listening to for internal communications.
  #     bind_addr: "0.0.0.0"
  #     bind_port: 5000

  #     # The address and port of a node that is already a member of the cluster that will help us join them.
  #     join_addr: "10.0.0.1"
  #     join_port: 5000

  #     # The address and port we provide to the to other peers to contact us.
  #     # This is rarely necessary, perhaps when behind NAT routing and the self IP detection discovery fails.
  #     advertise_addr: ""
  #     advertise_port: 5000

  #     # The maximum amount of time waiting for responses from other nodes in the cluster.
  #     relay_timeout: "5s"
```

Nodes form an overlay network and are able to route packets amongst themselves to work around NAT firewalls and other impediments. I personally use the Kubernetes DNS `tracker.namespace.svc.cluster.local` as the join address, all it takes is one known node to join the cluster.

# Performance

This is still incapable of running my tracker past 20% of the load, so I'm still investigating what's wrong.

Current metrics to give an idea of what 100% would mean:

153891872-e54e8dab-4eed-445b-8fe8-1f05dfefcee4

### DIFF
--- a/cmd/chihaya/config.go
+++ b/cmd/chihaya/config.go
@@ -18,6 +18,7 @@ import (
 	_ "github.com/chihaya/chihaya/middleware/varinterval"
 
 	// Imports to register storage drivers.
+	_ "github.com/chihaya/chihaya/storage/cluster"
 	_ "github.com/chihaya/chihaya/storage/memory"
 	_ "github.com/chihaya/chihaya/storage/redis"
 )
--- a/cmd/chihaya/signal_windows.go
+++ b/cmd/chihaya/signal_windows.go
@@ -5,7 +5,6 @@ package main
 
 import (
 	"os"
-	"os/signal"
 	"syscall"
 )
 
--- a/dist/example_config.yaml
+++ b/dist/example_config.yaml
@@ -111,7 +111,6 @@ chihaya:
     # The maximum number of infohashes that can be scraped in one request.
     max_scrape_infohashes: 50
 
-
   # This block defines configuration used for the storage of peer data.
   storage:
     name: "memory"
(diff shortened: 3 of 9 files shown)


mrd0ll4r:
Sorry for the slow response.
I had a look at this and thought about it for a bit. We thought about something like this some years ago, but ended up with: Probably easier to write a thin loadbalancer in Rust (or something else that's fast) that forwards requests to some Chihaya instances.

This should be trivial and very fast for UDP announces, since the infohash is at a constant offset in the datagram.
As an added bonus: The algorithm Chihaya uses to compute transaction IDs for UDP is consistent between different instances with a synchronized(-ish) clock and the same key. That means you could run extra instances just for transaction ID generation, and send connect requests to those.

For HTTP this entails parsing the HTTP request and query, which is duplicated work in the end, but should (hopefully) still be fine.
The loadbalancer would then use long-living HTTP/2 connections with the Chihaya instances and forward requests that way.

Thoughts?

nitrix:
Honestly would be better as its own load balancer, especially if it can plug into Kubernetes's API to get notified of new pods addition/removal/health checks; it would be very solid with zero to none unserviceable request.

Another option is Chihaya + Twemproxy + Redis, which I haven't explored yet.

Myeah, I wrote this out of ignorance really :P

jzelinskie:
>especially if it can plug into Kubernetes's API to get notified of new pods addition/removal/health checks

We use [kuberesolver](https://github.com/sercand/kuberesolver) for [SpiceDB](https://github.com/authzed/spicedb) (my day job) for exactly this. It's quite minimal code to get that list and then it's just a matter of load balancing with a consistent hash based on the connection ID.

PR #567, opened 2022-04-14
*: migrate to netip

jzelinskie:
deprecates #540


storage/redis/peer_store.go
```diff
@@ -785,16 +779,16 @@ func (ps *peerStore) collectGarbage(cutoff time.Time) error {
 				//_, err := conn.Do("DEL", ihStr)
 
 				_ = conn.Send("MULTI")
-				_ = conn.Send("HDEL", group, ihStr)
+				_ = conn.Send("HDEL", "I"+af, ihStr)
 				if isSeeder {
-					_ = conn.Send("DECR", ps.infohashCountKey(group))
+					_ = conn.Send("DECR", "I"+af)
```
mrd0ll4r, line 784: Something about these few lines seems wrong. We call `HDEL` and `DECR` on the same key, which I find pretty dubious.
Tbh I don't have the structure of all this stuff in redis in mind, so I don't know which one is correct, but one of them is probably not correct :)

mrd0ll4r (changes requested):
Nice! I'm mostly okay with this. Something about the redis implementation seems odd, and I don't remember how exactly that was laid out.
Now that I think about it, wasn't there also some markdown doc about that, which should maybe be updated with the new redis keys?

storage/redis/peer_store.go
```diff
@@ -269,28 +239,58 @@ type peerStore struct {
...
-}
+// Key that stores the peers for an InfoHash.
+//
+// "4S0102030405060708090a0b0c0d0e0f1011121314" => IPv4 Seeder with InfoHash 0102030405060708090a0b0c0d0e0f1011121314
+// "6L0204090405060708060a0b0c0d0e0f1011121328" => IPv6 Leecher with Infohash 0204090405060708060a0b0c0d0e0f1011121328
+func swarmKey(infoHash bittorrent.InfoHash, seed bool, addr netip.Addr) string {
+	var b strings.Builder
+	b.Grow(1 + 1 + 20) // "4"/"6" + "S"/"L" + len(InfoHash)
+
+	if addr.Is4() {
```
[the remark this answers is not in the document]
jzelinskie: Yeah, I haven't quite landed how I feel about where we _should_ be calling `addr.Unmap()` and then just assume everywhere else in the codebase that an IP cannot be 4in6.

storage/redis/peer_store.go
```diff
@@ -269,28 +239,58 @@ type peerStore struct {
...
 
-func (ps *peerStore) seederCountKey(af string) string {
-	return af + "_S_count"
+// Key that stores the cardinality of the peers of an IP version.
+//
+// "4L" => IPv4 Leechers
+// "6S" => IPv6 Seeders
+func peerCountKey(seed bool, addr netip.Addr) string {
+	key := "4"
+	if addr.Is6() {
```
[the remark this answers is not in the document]
jzelinskie: We call `addr.Unmap()` in Sanitize announce so there really should never be any time that the storage systems have to deal with 4in6.

storage/redis/peer_store.go
```diff
@@ -350,32 +350,32 @@ func (ps *peerStore) PutSeeder(ih bittorrent.InfoHash, p bittorrent.Peer) error
 	default:
 	}
 
-	pk := newPeerKey(p)
-
-	encodedSeederInfoHash := ps.seederInfohashKey(addressFamily, ih.String())
-	ct := ps.getClock()
+	addr := p.AddrPort.Addr()
+	sk := swarmKey(ih, true, addr)
+	ihKey := sk[0]
```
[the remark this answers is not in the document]
jzelinskie: Yeah, this confused me a bit too looking back. I'm open to suggestions because it's non obvious to me, haha.

storage/redis/peer_store.go
jzelinskie: This was definitely a bug. I found another one as well, so I created an addressFamily type that helps build these keys. All the methods are small enough that they should be inlined, so zero-cost.

mrd0ll4r:
(replying to this via email, hope that works)Yeah, that makes sense. I've looked this up for another project a while back: these 4-in-6 constructions were added for backwards compatibility: for applications that only support IPv6, on hosts that are dual stacked or IPv4, the OS can wrap those into IPv6. Maybe that also applies if we bind to IPv6 only and some netfilter stuff translates IPv4 to that. All in all probably somewhat unlikely... Maybe we should have never opened this can of worms 😅Anyway. Sanitizing in the sanitization Middleware seems like the sane thing. On Apr 16, 2022 18:47, Jimmy Zelinskie ***@***.***> wrote:
@jzelinskie commented on this pull request.



In storage/redis/peer_store.go:
>  }
 
-func (ps *peerStore) seederCountKey(af string) string {
-	return af + "_S_count"
+// Key that stores the cardinality of the peers of an IP version.
+//
+// "4L" => IPv4 Leechers
+// "6S" => IPv6 Seeders
+func peerCountKey(seed bool, addr netip.Addr) string {
+	key := "4"
+	if addr.Is6() {

We call addr.Unmap() in SanitizeAnnounce so there really should never be any time that the storage systems have to deal with 4in6.

jzelinskie:
I think this is good to go now @mrd0ll4r. I'd like to create some follow ups before I add too many changes to this one PR.

frontend/http/writer.go
```diff
@@ -37,22 +37,22 @@ func WriteAnnounceResponse(w http.ResponseWriter, resp *bittorrent.AnnounceRespo
...
+			ipv4CompactDict = append(ipv4CompactDict, compact4(peer)...)
 		}
-		if len(IPv4CompactDict) > 0 {
-			bdict["peers"] = IPv4CompactDict
+		if len(ipv4CompactDict) > 0 {
+			bdict["peers"] = ipv4CompactDict
 		}
 
 		// Add the IPv6 peers to the dictionary.
+		ipv6CompactDict := make([]byte, 0, compact6PeerLen*len(resp.IPv4Peers))
```
sot-tech, line 50: Did you mean `len(IPv6Peers)`?

jzelinskie: absolutely! haha autocomplete got me!

.github/workflows/lint.yaml
```diff
@@ -1,70 +1,33 @@
...
-      - uses: "golangci/golangci-lint-action@v2"
+      - uses: "actions/checkout@v3"
+      - uses: "actions/setup-go@v3"
         with:
-          version: "v1.43"
-          skip-go-installation: true
-          skip-pkg-cache: true
-          skip-build-cache: false
+          go-version: "^1.18"
+      - uses: "authzed/actions/gofumpt@main"
```
mrd0ll4r, line 19: I will steal your github actions for my personal stuff :)

storage/redis/peer_store.go
```diff
@@ -581,38 +586,38 @@ func (ps *peerStore) AnnouncePeers(ih bittorrent.InfoHash, seeder bool, numWant
 
 	if seeder {
 		// Append leechers as possible.
-		for _, pk := range conLeechers {
+		for _, sk := range conLeechers {
```
mrd0ll4r, line 589: I'd prefer these to stay `pk` as in peerKey, since they are encoded peers.

mrd0ll4r (changes requested):
Alright. A few more small ones and one more relevant one about sanitization of scrapes.
The redis stuff is still unclear in my head. I hope the tests catch the bugs...

middleware/logic_test.go
```diff
@@ -42,12 +42,16 @@ func (hooks hookList) handleAnnounce(ctx context.Context, req *bittorrent.Announ
...
+	req := &bittorrent.AnnounceRequest{Peer: bittorrent.Peer{
+		AddrPort: netip.AddrPortFrom(netip.MustParseAddr("1.2.3.4"), 0),
+	}}
 	benchHookList(b, hooks, req)
 }
 
 func benchHookListV6(b *testing.B, hooks hookList) {
-	req := &bittorrent.AnnounceRequest{Peer: bittorrent.Peer{IP: bittorrent.IP{IP: net.ParseIP("fc00::0001"), AddressFamily: bittorrent.IPv6}}}
+	req := &bittorrent.AnnounceRequest{Peer: bittorrent.Peer{
+		AddrPort: netip.AddrPortFrom(netip.MustParseAddr("fc00:0001"), 0),
```
sot-tech, line 53: `netip.MustParseAddr` will panic here. You forgot second colon in ipv6 address.

PR #574, opened 2022-05-25
Expand explanations on docker compose, systemd and default confs

shyba:


shyba:
ops, sorry. Wrong repo. Those changes are for running the service locally here

PR #596, opened 2022-10-06
ban xunlei

johnrosen1:
https://tsingjyujing.github.io/blog/other-tech/fuck-thunder.html

https://zhuanlan.zhihu.com/p/87193566

notorious leech only bittorrent client , should be banned by default


jzelinskie:
I'm not sure that we want to assume any particular use case for the defaults in chihaya.

In addition to users running chihaya on the open internet, there are quite a few folks running it on private networks internally at businesses. They are unlikely to care about the client approval middleware or this particular client.

WDYT @mrd0ll4r?

Issue #569, opened 2022-04-25
Change storage interface to take a slice of peers as input


shyba:
From a [previous discussion](https://github.com/chihaya/chihaya/pull/566#issuecomment-1098791659) on #566, @mrd0ll4r commented:

> (...)Thinking about it some more, the most elegant approach would be to pass a (possibly non-empty) slice of bittorrent.Peer to the storage to fill up to numWant, but we can't do that with the current interfaces. (...)

The relevant interface part is [AnnouncePeers, from storage.go](https://github.com/chihaya/chihaya/blob/main/storage/storage.go#L95), with updates to everything implementing it ([Redis](https://github.com/chihaya/chihaya/tree/main/storage/redis) and [memory](https://github.com/chihaya/chihaya/tree/main/storage/memory)).

PR #566, opened 2022-04-13
middleware/hooks: preserve prehook response changes

shyba:
Hi, thanks for this amazing project.

While implementing a prehook middleware, my changes to IPv4Peers were being discarded. This is what I had to do in order to fix it.

(I don't know if it is expected for prehook middleware implementations to modify responses, so feel free to discard the PR as well. The middleware I am working on needs to inject peers and using this hook was the cleanest approach.)

PR #602, opened 2023-01-31
storage: modify interface to pass a slice on AnnouncePeers

shyba:
This is an attempt at #569 

It has been a while since that was proposed, so idk if this is implemented the way we discussed in #566 but I'm happy to change


mrd0ll4r:
Hey, thanks! I think the idiomatic Go way would be to pass in a slice (by value) and have the function return a slice, similar to how `append` works.

And in the future, though that is for another PR, we can then pre-allocate (to something like MaxNumWant) and reuse these slices. We could also do that with the pointer solution, don't get me wrong, I just think the slice-by-value solution is more "natural" for Go.
And then, ultimately, we would limit the number of Goroutines (similarly to what you did in the other PR), and each of them has all the state necessary to serve a request pre-allocated. We probably won't get to zero allocations per request, because we still need to store the peer somewhere, but we should get it down by a lot.

As you can see, CI is still broken... If you're really motivated, you could try to figure out why that is and get it to work again.
There's a bunch of other PRs that should probably be merged first:
- https://github.com/chihaya/chihaya/pull/567 because it changes the network address types, which should be useful
- Maybe leftovers from https://github.com/chihaya/chihaya/pull/540 after #567 is merged, not sure if there are other things in there
Both of those are big and change fundamental things about the codebase. Also apparently I wasn't entirely happy with them yet, so they'd need some work first...

Or maybe we'll merge the small ones first and then change the big ones, what do I know.

Thanks again!

shyba:
Nice, that sounds like a very good plan.

I will change that to be idiomatic and take a look on CI.

shyba:
It now does pass by value instead of using a pointer. This is probably blocked by the other PRs that should go first, so I will explore that optimization of reusing the slice on the handler while that goes on.

storage/storage.go
```diff
@@ -93,7 +93,7 @@ type PeerStore interface {
 	//   leechers
 	//
 	// Returns ErrResourceDoesNotExist if the provided InfoHash is not tracked.
-	AnnouncePeers(infoHash bittorrent.InfoHash, seeder bool, numWant int, p bittorrent.Peer) (peers []bittorrent.Peer, err error)
+	AnnouncePeers(infoHash bittorrent.InfoHash, seeder bool, numWant int, p bittorrent.Peer, peers []bittorrent.Peer) ([]bittorrent.Peer, error)
```
mrd0ll4r, line 96: This could use an update to the godoc stating what happens for error cases. You opted to return the (possibly already modified) peer slice together with the error. I think that's fine, but it needs to be documented. We could alternatively also document that only `nil, <some error>` or `<some peer slice>, nil` may be returned. In that case, we'd need to change the implementations.
The idiomatic Go approach would be to return `nil, <some error>` on error. I'd be fine with your approach, though.

storage/storage_bench.go
```diff
@@ -405,7 +405,7 @@ func AnnounceLeecher(b *testing.B, ps PeerStore) {
 // AnnounceLeecher1kInfohash can run in parallel.
 func AnnounceLeecher1kInfohash(b *testing.B, ps PeerStore) {
 	runBenchmark(b, ps, true, putPeers, func(i int, ps PeerStore, bd *benchData) error {
-		_, err := ps.AnnouncePeers(bd.infohashes[i%1000], false, 50, bd.peers[0])
```
mrd0ll4r, line 408: There's a reason this was `bd.infohashes[i%1000]`, namely that those are different infohashes, which are potentially stored in different shards of the storage, which allows this benchmark to run in parallel.

storage/storage_bench.go
```diff
@@ -427,7 +427,7 @@ func AnnounceSeeder(b *testing.B, ps PeerStore) {
 // AnnounceSeeder1kInfohash can run in parallel.
 func AnnounceSeeder1kInfohash(b *testing.B, ps PeerStore) {
 	runBenchmark(b, ps, true, putPeers, func(i int, ps PeerStore, bd *benchData) error {
-		_, err := ps.AnnouncePeers(bd.infohashes[i%1000], true, 50, bd.peers[0])
```
mrd0ll4r, line 430: Same here

storage/storage_bench.go
```diff
@@ -416,7 +416,7 @@ func AnnounceLeecher1kInfohash(b *testing.B, ps PeerStore) {
 // AnnounceSeeder can run in parallel.
 func AnnounceSeeder(b *testing.B, ps PeerStore) {
 	runBenchmark(b, ps, true, putPeers, func(i int, ps PeerStore, bd *benchData) error {
-		_, err := ps.AnnouncePeers(bd.infohashes[0], true, 50, bd.peers[0])
+		_, err := ps.AnnouncePeers(bd.infohashes[0], false, 50, bd.peers[0], nil)
```
mrd0ll4r, line 419: And this needs `true` as the second parameter, otherwise it tests AnnounceLeecher.

storage/storage_tests.go
```diff
@@ -47,7 +47,7 @@ func TestPeerStore(t *testing.T, p PeerStore) {
 		err = p.DeleteSeeder(c.ih, c.peer)
 		require.Equal(t, ErrResourceDoesNotExist, err)
 
-		_, err = p.AnnouncePeers(c.ih, false, 50, peer)
+		_, err = p.AnnouncePeers(c.ih, true, 50, peer, nil)
```
mrd0ll4r, line 50: Same here.

storage/storage_tests.go
```diff
@@ -86,7 +86,7 @@ func TestPeerStore(t *testing.T, p PeerStore) {
 		err = p.DeleteLeecher(c.ih, c.peer)
 		require.Nil(t, err)
 
-		peers, err = p.AnnouncePeers(c.ih, true, 50, peer)
+		peers, err = p.AnnouncePeers(c.ih, false, 50, peer, nil)
```
mrd0ll4r, line 89: Same here

mrd0ll4r (changes requested):

storage/storage.go
shyba: Good point. I didn't know that. Changed back to `nil, peers` as that sounds more idiomatic and added a line explaining the new optional parameter.

PR #607, opened 2023-02-05
Fix to HTTP header X-Forwarded-For parse to support proxy chains.

nitrix:
Hello,

Ingress traffic going through a proxy chain leads to `X-Forwarded-For` headers that are comma-separated enumerations.
The first item in the list is the client, every one else are the participating proxy forming the chain of trust.

This small change picks the only IP / or the first IP of the list to avoid failing with a parse error.

It resolved the issue on my side for my open proxy and would love to see it merged upstream.

Cheers,
Alex.

### DIFF
--- a/frontend/http/parser.go
+++ b/frontend/http/parser.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"net"
 	"net/http"
+	"strings"
 
 	"github.com/chihaya/chihaya/bittorrent"
 )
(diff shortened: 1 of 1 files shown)


mrd0ll4r:
Hello and thank you :)

> Ingress traffic going through a proxy chain leads to X-Forwarded-For headers that are comma-separated enumerations.

Is this official behavior, documented somewhere?

nitrix:
@mrd0ll4r Sort of, X-types of headers are by consensus and not part of the HTTP RFCs but I imagine there's one somewhere that talks about it. Mozilla has a great explanation: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#syntax

I would say though, don't merge this PR yet; the port support (which is kind of very esoteric) isn't playing nice with IPv6 addresses. I'll get back to you with a nicer proper implementation.

PR #618, opened 2023-05-01
frontend/udp: ensure race-free shutdown

### LINKED ISSUE #608: Data Race in UDP Frontend
The data race in the UDP Frontend seems to be because any combination of these:

1. we call `(*sync.WaitGroup).Add` from a different goroutine than the one calling `(*sync.WaitGroup).Wait`, e.g., [here](https://github.com/chihaya/chihaya/blob/main/frontend/udp/frontend.go#L174) and [here](https://github.com/chihaya/chihaya/blob/main/frontend/udp/frontend.go#L205)
2. we call `(*sync.WaitGroup).Add` _after_ a call to `(*sync.WaitGroup).Wait` has started, i.e., [here](https://github.com/chihaya/chihaya/blob/main/frontend/udp/frontend.go#L205). This seems more likely to me. We set the read deadline on the socket and call `Wait`, but `serve()` still has some packet data and launches a goroutine to process that.

I also noticed that `(t *Frontend).Stop()` is not thread-safe by my definition, because it could close `t.closing` multiple times, which would panic.

In general, that needs a small rework with
1. A mutex to govern write-access to `closing`
2. A signalling/synchronization mechanism that ensures we will never call `(*sync.WaitGroup).Add` after we set the socket read deadline (or some other useful point)
3. Probably moving [this](https://github.com/chihaya/chihaya/blob/main/frontend/udp/frontend.go#L174) outside of the goroutine, because it really is a bit weird to have it in there.

mrd0ll4r:
Fixes #608 

New mechanism to shut down cleanly:
1. `closing` is closed, only once, guarded by the mutex.
2. Socket read deadline is set, which ensures that `serve` eventually restarts its loop and realizes we're closing.
3. `serve` closes `serveShutdown` when it returns.
4. `serveShutdown` signals `Stop` that `serve` has returned and no more calls to `f.wg.Add` can be made.
`5. f.wg.Wait` is called, which should be race-free now.

Hope that make sense.
Also, some cleanup with where `f.wg.Add` is called.
Also, this should hopefully finally fix the flaky CI and allow us to merge a bunch of PRs, maybe.

### DIFF
--- a/frontend/udp/frontend.go
+++ b/frontend/udp/frontend.go
@@ -98,9 +98,18 @@ func (cfg Config) Validate() Config {
 
 // Frontend holds the state of a UDP BitTorrent Frontend.
 type Frontend struct {
-	socket  *net.UDPConn
+	socket *net.UDPConn
+
+	// A mutex for the closing channel, to make sure concurrent calls to Stop
+	// are thread-safe.
+	closingM sync.Mutex
+	// A channel to indicate we are shutting down.
 	closing chan struct{}
-	wg      sync.WaitGroup
+	// A channel that is closed once serve exits. Ensures clean shutdown.
+	serveShutdown chan struct{}
+	// A wait group for all goroutines spawned by the frontend.
+	// After serveShutdown is closed, no more calls to Add can be made.
+	wg sync.WaitGroup
 
 	genPool *sync.Pool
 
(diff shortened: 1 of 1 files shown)


mrd0ll4r:
@jzelinskie 👉 👈

jzelinskie (approved):

PR #623, opened 2026-06-21
the great a-makening

jzelinskie:
Updates *everything* and adopts a Makefile.

This monster commit updates:
- Go toolchain
- All dependencies
- All linters
- GitHub Actions

Because GHA require pinned SHAs for safety, it is perpetually out of date. Thus, this commit moves as much as possible into a Makefile to avoid any unnecessary actions. This has the added benefit of making it so that mostly everything ran in CI can be easily reproduced locally.

I don't like Makefiles, but GHA is ridiculous.

This also adopts the new `go tool` command for dev tooling where possible. redis-server becomes the only dependency that isn't managed.


bittorrent/sanitize.go
```diff
@@ -25,11 +25,11 @@ func SanitizeAnnounce(r *AnnounceRequest, maxNumWant, defaultNumWant uint32) err
...
 
-	if ip := r.Peer.IP.To4(); ip != nil {
-		r.Peer.IP.IP = ip
-		r.Peer.IP.AddressFamily = IPv4
-	} else if len(r.Peer.IP.IP) == net.IPv6len { // implies r.Peer.IP.To4() == nil
-		r.Peer.IP.AddressFamily = IPv6
+	if ip := r.IP.To4(); ip != nil {
+		r.IP.IP = ip
+		r.IP.AddressFamily = IPv4
+	} else if len(r.IP.IP) == net.IPv6len { // implies r.Peer.IP.To4() == nil
```
mrd0ll4r: The comment still references `b.Peer.IP`

frontend/http/writer.go
```diff
@@ -37,9 +37,8 @@ func WriteAnnounceResponse(w http.ResponseWriter, resp *bittorrent.AnnounceRespo
 
 	// Add the peers to the dictionary in the compact format.
 	if resp.Compact {
-		var IPv4CompactDict, IPv6CompactDict []byte
-
 		// Add the IPv4 peers to the dictionary.
+		IPv4CompactDict := make([]byte, 0, 8*len(resp.IPv4Peers))
```
mrd0ll4r: I'd have expected `6*len(resp.IPv4Peers`, `4` bytes for the IP address and `2` bytes for the port?

.yamlfmt.yaml
```diff
@@ -0,0 +1,2 @@
+exclude:
+  - "dist"
```
mrd0ll4r: Why are these excluded?

go.mod
```diff
@@ -1,25 +1,262 @@
 module github.com/chihaya/chihaya
 
-go 1.16
+go 1.25.0
+
+toolchain go1.26
```
mrd0ll4r, line 5: No clue how recent `go.mod` files work, but the toolchain version is higher than the go version itself, is this correct?

mrd0ll4r (changes requested):
Nice! Also not a fan of GHA, but mostly due to the arcane syntax (and impossibility to test locally, I guess). I'm not convinced this makes it so that we can run the same CI locally -- e.g. the `go tool` calls, do those not need pinned versions to be reproducible? Then again, I agree that the permanent dependabot and version-bumping spam is pretty annoying.

Something I'd like to see is a paragraph in the README that talks about the Makefile.

Also, I thought about moving e2e tests to Docker at some point, or even the entire CI. I've been doing that with my Rust projects to have it compile the same way everywhere. But more something to discuss than to implement now :)

I added a few comments here and there, nothing major.

go.mod
jzelinskie: I have to constantly reread how it works, but IIRC, `go` is the min version and toolchain is the one for reproducible builds. So you basically let `go get` figure out the `go` one and then you set the toolchain to your own preferences. Our `setup-go` GHA is reading the go.mod and picking 1.26 because that's the toolchain version.

.yamlfmt.yaml
jzelinskie: I actually meant to just exclude helm because it is templated YAML. I'll fix that up.

frontend/http/writer.go
jzelinskie: Totally just fat-fingered this one. Nice catch!

jzelinskie:
> I'm not convinced this makes it so that we can run the same CI locally -- e.g. the go tool calls, do those not need pinned versions to be reproducible? Then again, I agree that the permanent dependabot and version-bumping spam is pretty annoying.

I actually first naively tried to add a version, but apparently tools don't support pinning to a version (crazy, right?).
I think this is still a step up from the current state of affairs.

>Also, I thought about moving e2e tests to Docker at some point, or even the entire CI. I've been doing that with my Rust projects to have it compile the same way everywhere. But more something to discuss than to implement now :)

Using test containers? I'm not sure it makes much of a difference here because chihaya is static binary.

mrd0ll4r (approved):
LGTM, thanks :)

mrd0ll4r:
> Using test containers? I'm not sure it makes much of a difference here because chihaya is static binary.

I usually build in one container with a stable base OS, e.g. Debian, and then extract the (statically linked) binary from the container.
But I was more thinking about the tools and dependencies we need for testing -- the e2e tests need Redis, not sure if there were ever any breaking changes in that, but we could pin it to something like "whatever the latest stable Debian uses" and outsource that problem, if that makes sense. Same with `helm` I guess.

The whole Go linting situation on the other hand is just not fixable, I feel like.
Whenever I touch one of my older Go projects to update stuff, or even just try to merge dependabot PRs,
some weird issue about linter versions, new lints, or some other blocker pops up.

PR #625, opened 2026-06-23
refactor: replace pkg/log & logrus with slog

jzelinskie:
This PR replaces logrus with slog + a very tiny library to add similar formatting.

slog is a standard Go library that should make it nicer for anyone importing chihaya into their own codebase with different logging configured.

slog should also allocate less than logrus.

I've opted for using the syntax for logger that allocates the least rather than the ones that are syntacticly the most ergonomic.


bittorrent/bittorrent.go
```diff
@@ -136,16 +135,27 @@ type AnnounceResponse struct {
...
 	}
+
+	ipv6s := make([]slog.Value, 0, len(r.IPv6Peers))
+	for _, p := range r.IPv6Peers {
+		ipv6s = append(ipv6s, p.LogValue())
+	}
+
+	return slog.GroupValue(
+		slog.Bool("compact", r.Compact),
+		slog.Uint64("complete", uint64(r.Complete)),
```
mrd0ll4r, line 153: This is missing `incomplete`. The original `LogFields` was also missing that, but I don't see a reason why 🤔

cmd/chihaya/main.go
```diff
@@ -51,16 +53,16 @@ func (r *Run) Start(ps storage.PeerStore) error {
 
 	r.sg = stop.NewGroup()
 
-	log.Info("starting metrics server", log.Fields{"addr": cfg.MetricsAddr})
+	slog.Info("starting metrics server", slog.String("addr", cfg.MetricsAddr))
 	r.sg.Add(metrics.NewServer(cfg.MetricsAddr))
 
 	if ps == nil {
-		log.Info("starting storage", log.Fields{"name": cfg.Storage.Name})
+		slog.Info("starting storage", "name", cfg.Storage.Name)
```
mrd0ll4r: I'm confused why we sometimes use `slog.Info(..., slog.String(key, value))` and `slog.Info(..., key, value)` at other times.
I looked into the implementation a bit:
- If we call `slog.String`, that (potentially?) allocates an `Attr` regardless of whether logging is enabled at that level
- If we provide key/value pairs, it does not allocate if the level is disabled (https://cs.opensource.google/go/go/+/refs/tags/go1.26.4:src/log/slog/logger.go;l=244) but later creates an `AnyValue` ad-hoc from the key and the next element (https://cs.opensource.google/go/go/+/refs/tags/go1.26.4:src/log/slog/record.go;l=174), which is likely more expensive because it needs to figure out the type.

I guess the `slog.Info(..., key, value)` variant is better then?
On the other hand, I don't know how `... any` args are implemented under the hood and whether providing (too) many of them is expensive too.

Opinions?

frontend/http/frontend.go
```diff
@@ -35,26 +36,26 @@ type Config struct {
...
+// LogValue renders a config as a set of log fields.
+func (cfg Config) LogValue() slog.Value {
+	return slog.GroupValue(
+		slog.String("addr", cfg.Addr),
+		slog.String("httpsAddr", cfg.HTTPSAddr),
+		slog.Duration("readTimeout", cfg.ReadTimeout),
+		slog.Duration("writeTimeout", cfg.WriteTimeout),
+		slog.Duration("idleTimeout", cfg.IdleTimeout),
+		slog.Bool("enableKeepAlive", cfg.EnableKeepAlive),
+		slog.String("tlsCertpath", cfg.TLSCertPath),
```
mrd0ll4r: The capitalization changed

frontend/http/frontend.go
```diff
@@ -35,26 +36,26 @@ type Config struct {
...
+		slog.String("tlsKeyPath", cfg.TLSKeyPath),
+		slog.Any("announceRoutes", cfg.AnnounceRoutes),
+		slog.Any("scrapeRoutes", cfg.ScrapeRoutes),
+		slog.Bool("enableRequestTiming", cfg.EnableRequestTiming),
+		slog.Bool("allowIPSpoofing", cfg.AllowIPSpoofing),
```
mrd0ll4r: As an alternative, would it make sense to implement `LogValue` for `ParseOptions` and include that here as a group or something? The mental overhead of maintaining these here seems higher.

frontend/http/frontend.go
```diff
@@ -73,60 +74,66 @@ func (cfg Config) Validate() Config {
...
 	if cfg.ReadTimeout <= 0 {
 		validcfg.ReadTimeout = defaultReadTimeout
-		log.Warn("falling back to default configuration", log.Fields{
-			"name":     "http.ReadTimeout",
-			"provided": cfg.ReadTimeout,
-			"default":  validcfg.ReadTimeout,
-		})
+		slog.Warn(
+			"falling back to default configuration",
+			slog.String("name", "http.ReadTimeout"),
```
mrd0ll4r, line 75: Same as in `main`: maybe move to name/value pairs instead of building the `Attr`s here?
(On the other hand, this is not the hot path, so meh.)

bittorrent/bittorrent.go
```diff
@@ -136,16 +135,27 @@ type AnnounceResponse struct {
...
+		slog.Duration("interval", r.Interval),
+		slog.Duration("minInterval", r.MinInterval),
+		// TODO(jzelinskie): avoid reflection for these
```
mrd0ll4r, line 157: Hmm, I had a look at the `slog` API but couldn't figure out a better way...

storage/redis/peer_store.go
```diff
@@ -339,10 +356,11 @@ func (ps *peerStore) getClock() int64 {
...
 func (ps *peerStore) PutSeeder(ih bittorrent.InfoHash, p bittorrent.Peer) error {
 	addressFamily := p.IP.AddressFamily.String()
-	log.Debug("storage: PutSeeder", log.Fields{
-		"InfoHash": ih.String(),
-		"Peer":     p,
-	})
+	slog.Debug(
+		"storage: PutSeeder",
+		slog.String("infoHash", ih.String()),
+		slog.Attr{Key: "peer", Value: p.LogValue()},
```
mrd0ll4r: I believe we should always avoid calling `LogValue()` ourselves to defer this as much as possible, no?

mrd0ll4r (changes requested):
Generally a fan of this, but:

- We need to figure out whether to do `slog.Info("foo", slog.String("key", "value"))` vs `slog.Info("foo", "key", "value")`. See the discussion in one of the comments. I think the latter is preferable to defer allocations as much as possible, i.e., after `level` is checked. We lose out on some static typing, and incur runtime cost because it has to check `"value".(type)`, but it should be faster if that level of logging is disabled. I'd at least prefer this for any logging on the hot path.
- Similarly, we should probably not call `foo.LogValue()` ourselves, but rather pass the thing to be logged (by reference, probably?). The `slog` docs make it look like this should work and is the intended way.
- Something about the embedded fields. I don't like that we create fields in the embedding struct's `LogValue`, that feels like it'd be more difficult to maintain.
- Personally not a big fan of colorized logs, call me a purist :) But also not strongly opposed.

jzelinskie:
>We need to figure out whether to do slog.Info("foo", slog.String("key", "value")) vs slog.Info("foo", "key", "value"). See the discussion in one of the comments. I think the latter is preferable to defer allocations as much as possible, i.e., after level is checked. We lose out on some static typing, and incur runtime cost because it has to check "value".(type), but it should be faster if that level of logging is disabled. I'd at least prefer this for any logging on the hot path.

This is generally the approach I took. There are only a few places I use the former and they are for setup events and not in the line of handling requests. That being said, it might be easier to just say "you must always do the performant way" for consistency's sake.

>Similarly, we should probably not call foo.LogValue() ourselves, but rather pass the thing to be logged (by reference, probably?). The slog docs make it look like this should work and is the intended way.

Yeah, tbh it was unclear what the most performant way to render the implementers of `slog.LogValuer`. I've now learned that you basically have to do this:

```go
ctx := context.Background()
if logger.Enabled(ctx, slog.LevelDebug) {
    logger.LogAttrs(ctx, slog.LevelDebug, "message",
        slog.String("key", value),
        slog.Int("count", n),
    )
}
```

>Something about the embedded fields. I don't like that we create fields in the embedding struct's LogValue, that feels like it'd be more difficult to maintain.

The text formatter doesn't embed them: it does `parent.key=value`. JSON is the only place where it embeds, which I think seems fine. Honestly, I didn't plan for this style of embedding; it's just what seems to happen naturally from using the slog API.

>Personally not a big fan of colorized logs, call me a purist :) But also not strongly opposed.

I just maintained the pre-existing behavior. We've got `--no-colors` and we detect if it's an interactive tty and default to no colors when it is not. TBH, I find the new colors easier to read.

cmd/chihaya/main.go
jzelinskie: I went with the typed version and found a linter to enforce it.

I figured out the way to avoid any allocations for the hot path and applied it in the hot path of requests and gc.

bittorrent/bittorrent.go
jzelinskie: great catch

bittorrent/bittorrent.go
jzelinskie: I think I'll leave this here.

middleware/jwt/jwt.go
```diff
@@ -182,66 +182,91 @@ func (h *hook) HandleScrape(ctx context.Context, _ *bittorrent.ScrapeRequest, _
...
 	}
 
 	ihHex := hex.EncodeToString(ih[:])
 	if ihClaim, ok := claims.Get("infohash").(string); !ok || ihClaim != ihHex {
-		log.Debug("unequal or missing infohash when validating JWT", log.Fields{
-			"exists":  ok,
-			"claim":   ihClaim,
-			"request": ihHex,
-		})
+		slog.Debug(
```
mrd0ll4r: This is missing the `Enabled` construction

storage/redis/peer_store.go
```diff
@@ -223,7 +245,10 @@ func New(provided Config) (storage.PeerStore, error) {
 			case <-t.C:
 				before := time.Now()
 				ps.populateProm()
-				log.Debug("storage: populateProm() finished", log.Fields{"timeTaken": time.Since(before)})
+				slog.Debug(
```
mrd0ll4r, line 243: This could maybe use the `Enabled` construction, too. Prometheus reporting happens more frequently than GC, I think.
But TBH I probably wouldn't do it for either, as they're both infrequent enough for my taste :)

EDIT: I _do_ support the construction within the GC routine, because we debug log every peer we remove etc., and that's probably a lot of overhead otherwise.

mrd0ll4r (changes requested):
About embedding: I think what I said was confusing. I was arguing that we shouldn't spell out fields of embedded structs in the parent struct's `LogValue`, not that the output implementation formats them in some way. We're now including the embedded structs with `slog.Any("embedded_struct", &embedded_struct)`, so this is fixed :)

About performance: Well, it looks disgusting, but I guess there's no way around it. I found two places where we should/could add the `Enabled` check.

storage/redis/peer_store.go
jzelinskie: I think it's fine to leave consider the Prometheus ones not in the critical path. It's also got a call to time.Now() that's wasted if you aren't debug logging, so it's no big deal.

I consider the GC path critical because it has contention with modifying the peer storage.

middleware/jwt/jwt.go
jzelinskie: nice catch

mrd0ll4r:
All the things I commented on are fixed, but something else popped up, sorry! Something to discuss maybe:

Should we enforce to implement `LogValue` on pointer receivers only? (Exception: As part of an interface, as whatever the interface is implemented on should be fine, see below) I was thinking something like this:
I don't exactly recall how this works in Go, but e.g. [here](https://github.com/chihaya/chihaya/blob/slog/bittorrent/sanitize.go#L42) we log `slog.Any("request", r)`, where `r` is `*AnnounceRequest`. But `LogValue` is implented for the non-pointer receiver `AnnounceRequest` (see [here](https://github.com/chihaya/chihaya/blob/slog/bittorrent/bittorrent.go#L109)). I feel like Go creates a copy of `r` for the `slog.Any` call. (Which, of course, only happens when debug logging is enabled, I know.)

I noticed this because of the embedding: I saw that you changed the receiver for `frontend.udp.ParseOptions.LogValue` to a pointer, which I think is correct. It's embedded in `frontend.udp.Config` and we log it as `slog.Any("parseOptions", &cfg.ParseOptions)`, which I also think is correct.

I checked all the other places where we implement `LogValue`:
```bash
$ grep -rF ' LogValue' | grep -vF '//'
middleware/jwt/jwt.go:func (cfg Config) LogValue() slog.Value {
storage/memory/peer_store.go:func (cfg Config) LogValue() slog.Value {
storage/memory/peer_store.go:func (ps *peerStore) LogValue() slog.Value { return ps.cfg.LogValue() }
storage/redis/peer_store.go:func (cfg Config) LogValue() slog.Value {
storage/redis/peer_store.go:func (ps *peerStore) LogValue() slog.Value { return ps.cfg.LogValue() }
frontend/udp/frontend.go:func (cfg Config) LogValue() slog.Value {
frontend/udp/parser.go:func (opts *ParseOptions) LogValue() slog.Value {
frontend/http/frontend.go:func (cfg Config) LogValue() slog.Value {
frontend/http/parser.go:func (opts *ParseOptions) LogValue() slog.Value {
bittorrent/bittorrent.go:func (r AnnounceRequest) LogValue() slog.Value {
bittorrent/bittorrent.go:func (r AnnounceResponse) LogValue() slog.Value {
bittorrent/bittorrent.go:func (r ScrapeRequest) LogValue() slog.Value {
bittorrent/bittorrent.go:func (sr ScrapeResponse) LogValue() slog.Value {
bittorrent/bittorrent.go:func (s Scrape) LogValue() slog.Value {
bittorrent/bittorrent.go:func (p Peer) LogValue() slog.Value {
```

I think the relevant ones would be all the `Request`s and `Response`s, `Peer`, and `Scrape`. The `Config`s... sure, why not, but those are not on the hot path, ever. But while we're at it, we could do those, too.

There's no lint for this in `sloglint`, I already checked.

Opinions?

jzelinskie:
I think to avoid an automatic dereference, you need to implement it on the pointer. As far as slog is concerned, if you implement it on the pointer and forget to pass the parameter by reference, `slog.Any` will gladly print a pretty string of the struct using reflection rather than throw an error. We could create a `slogutil` package an implement a `slog.Inter(name string, valuer slog.LogValuer)` to enforce the interface. I'm really surprised that this type of function is not already in the slog package because I'm pretty sure it exists for other pre-existing loggers like zerolog.

jzelinskie:
I cherrypicked the parts of slog we use into its own package to enforce the patterns we want as much as possible. I also took it as an opportunity to clean up the configuration in main, too.

Did I go overboard?

pkg/slog/slog.go
```diff
@@ -0,0 +1,195 @@
...
+// needed, or to expand a single value into a sequence of components.
+type LogValuer = slog.LogValuer
+
+// Valuer returns an Attr for an implementer of slog.LogValuer.
+//
+// This is useful because it enforces the interface at compile-time rather than
+// relying on slog.Any which has additional behavior that can be triggered
+// accidentally.
+func Valuer(name string, valuer LogValuer) slog.Attr {
+	return slog.Any(name, valuer)
```
mrd0ll4r: Maybe overkill, but: I think if we call `value.LogValue()` here instead of passing along the interface we avoid one round of reflection. `slog.Any` calls `slog.AnyValue`, which does not check for `LogValuer` (see [here](https://cs.opensource.google/go/go/+/refs/tags/go1.26.4:src/log/slog/value.go;l=221)) (wtf?) and stores the passed value as `any` as a fallback and thus erases the type information we have here. Maybe.

frontend/http/frontend.go
```diff
@@ -35,26 +36,22 @@ type Config struct {
...
+		slog.Duration("readTimeout", cfg.ReadTimeout),
+		slog.Duration("writeTimeout", cfg.WriteTimeout),
+		slog.Duration("idleTimeout", cfg.IdleTimeout),
+		slog.Bool("enableKeepAlive", cfg.EnableKeepAlive),
+		slog.String("tlsCertPath", cfg.TLSCertPath),
+		slog.String("tlsKeyPath", cfg.TLSKeyPath),
+		slog.Any("announceRoutes", cfg.AnnounceRoutes),
+		slog.Any("scrapeRoutes", cfg.ScrapeRoutes),
+		slog.Bool("enableRequestTiming", cfg.EnableRequestTiming),
+		slog.Any("parseOptions", &cfg.ParseOptions),
```
mrd0ll4r: `slog.Valuer` 😎

pkg/slog/slog.go
```diff
@@ -0,0 +1,195 @@
...
+	"io"
+	"log/slog"
+	"time"
+
+	"github.com/lmittmann/tint"
+)
+
+// Names for common levels.
+//
+// Level numbers are inherently arbitrary,
```
mrd0ll4r: I was pretty confused by the godoc, I guess this is copied from `slog`? We could also just say "The same as in slog` or something instead of the whole rationale, might be easier to digest :)

storage/storage.go
```diff
@@ -110,9 +110,9 @@ type PeerStore interface {
 	// For more details see the documentation in the stop package.
 	stop.Stopper
 
-	// log.Fielder returns a loggable version of the data used to configure and
-	// operate a particular PeerStore.
-	log.Fielder
+	// log.LogValuer returns a loggable version of the data used to configure
```
mrd0ll4r: Ah, didn't notice before: The godoc still refers to `log` instead of `slog`.

mrd0ll4r (changes requested):
Hah, now we went from a custom `log` to a custom `slog` 🙃 
I'm not sure if we need all of the things we cherry-picked, e.g., the levels and types, and the `slog.Attr` constructors that are copied. Do we need to redeclare them for our interfaces to work? Otherwise I'd elect to leave them in `slog` and only cover the minimum of what we need. Not sure how annoying that gets with import names, if both are imported as `slog`.

jzelinskie:
My intent for wrapping as much as possible was to basically make it so that all we need to do is sanity check "is the import for slog right?" and that guarantees the contents of the file is logging performantly.

I think it'd be nice to not have our own log package, but even if we did, it's beneficial for us to use slog because it's theoretically a "standard logger" that has connectors to all of the other libraries.

mrd0ll4r:
> I think it'd be nice to not have our own log package, but even if we did, it's beneficial for us to use slog because it's theoretically a "standard logger" that has connectors to all of the other libraries.

Oh yeah, no worries, I'm fine with it :)

About the wrapping: Ah, that makes sense, yes. I can't fully gauge whether this might eventually impact external middleware or something, but we can add stuff then if needed.

mrd0ll4r (approved):
LGTM 🚀

